PR #7, opened 2024-01-30
Only list on localhost by default for fabric_api_server.py

RichardoC:
1.1.1.1 is owned by another company, so the example IP should probably either be a local ip only (to prevent external use) or 0.0.0.0 if the assumption is that the server should be public.


danielmiessler:
I hear you and I appreciate you, but I don't see this causing trouble on the internet, unless I'm missing something. It's not meant to be run on 1.1.1.1. It's meant to remind the person to change that IP to what they want it to be. But thank you. And let me know if I'm missing something and this can actually cause a problem.

PR #11, opened 2024-01-31
Delete api key

yanisldv:
You better use dotenv to keep your api key hidden


KieranCrossland:
💀

danielmiessler:
That's not a real key. That's an example for the server config, so people can set it up for themselves with their own values.

PR #17, opened 2024-02-01
Your commit message here

SamuelNii32:


SwiftCoderJoe:
The MIT license is a license in wide use across many software projects. There is no need to change it, especially in the possible face of legal scrutiny.

danielmiessler:
I took this from the official license, so I'm hesitant to alter it.

PR #28, opened 2024-02-03
Use ConfigParser to parse user config file

RyanBalfanz:
This fixes a bug caused by extra characters (e.g. newline) leading to an invalid `~/.config/fabric/.env` file (according to the current parsing logic, that is). A newline should be allowed yet causes an exception to be raised with a traceback pointing elsewhere in the code. Such a traceback could be difficult to fix for some users and turn them away from fabric rather quickly. Plus, sometimes text editors will automatically add the newline.

```shell
Traceback (most recent call last):
  File "/Users/ryan/fabric/client/fabric", line 68, in <module>
    standalone.sendMessage(text)
  File "/Users/ryan/fabric/client/utils.py", line 100, in sendMessage
    print(response)
          ^^^^^^^^
UnboundLocalError: cannot access local variable 'response' where it is not associated with a value
```

Separately, I would also argue that the code around the traceback needs some changes too, but this change does fix the issue I first encountered.

### DIFF
--- a/client/utils.py
+++ b/client/utils.py
@@ -1,20 +1,35 @@
-import requests
+import configparser
 import os
-from openai import OpenAI
-import pyperclip
 import sys
+from dataclasses import dataclass
+from typing import Self
+
+import pyperclip
+import requests
+from openai import OpenAI
 
 current_directory = os.path.dirname(os.path.realpath(__file__))
 config_directory = os.path.expanduser("~/.config/fabric")
-env_file = os.path.join(config_directory, '.env')
+env_file = os.path.join(config_directory, ".env")
+
+
+@dataclass
+class UserConfig:
+    openai_api_key: str
+
+    @classmethod
+    def from_env_file(cls, filepath: str) -> Self:
+        config = configparser.ConfigParser()
+        with open(env_file) as stream:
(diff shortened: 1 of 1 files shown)


RyanBalfanz:
A simpler change that would also fix the extra newline in the `.env` file would be to simply strip `apikey`. If that is preferred I can update this PR.

```python
apikey = f.read().split("=")[1].strip()
```

I didn't want to introduce any new dependencies (e.g. python-dotenv, pydantic-settings), preferred to use the standard library, hence `configparser`.

client/utils.py
```diff
@@ -1,20 +1,35 @@
-import requests
+import configparser
 import os
-from openai import OpenAI
-import pyperclip
 import sys
+from dataclasses import dataclass
+from typing import Self
```
RyanBalfanz, line 5: [New in version 3.11](https://docs.python.org/3/library/typing.html#typing.Self)

RyanBalfanz:
> A simpler change that would also fix the extra newline in the `.env` file would be to simply strip `apikey`. If that is preferred I can update this PR.

Alternative PR at https://github.com/danielmiessler/fabric/pull/29

danielmiessler:
I think we fixed this by pulling the key in a different way. Thank you so much for this, though!

PR #29, opened 2024-02-03
Strip the OpenAI API Key when parsing the .env config file

RyanBalfanz:
Alternative to #28


danielmiessler:
I think we fixed this, thank you for the comment!

Issue #27, opened 2024-02-02
Use python poetry for managing dependencies in a deterministic fashion


agu3rra:
I noticed the repo contains a collection of `requirements.txt` file without pinned dependencies. This translates to every time a "build" (dependencies pull in the case of python) happens, a new set of dependencies will be resolved (time dependent), which could lead to issues reproducing the same exact dev environment on multiple machines.

I suggest using https://python-poetry.org/ to manage dependencies as it is really efficient and makes it clear which dependencies you rely on.

I volunteer to submit a PR if you're up for it. Just let me know @danielmiessler.

Cheers! \o/

PS: I assume you wish to manage the `client` and `server` dependencies separately. Please confirm if that's the case.

AvikantSrivastava:
@agu3rra @danielmiessler 
can I work on this?

I have experience in setting up poetry from scratch on a codebase

danielmiessler:
This sounds good but it's a bit beyond my Python dependency skills. Can you describe what all changes this would entail?

The prime directive of the project's codebase is simplicity. So if we can do this without adding complexity I'm happy to hear about it.

agu3rra:
Hello @danielmiessler 
I've submitted the above PR #32 to show you what it will look like. I have a hunch the `sources` folder may not be required, but that was just a hunch. Let me know if not.

The PR has a screenshot of how the install and CLI command will look like.

I strive for simplicity and good laziness :)

martin-luecke:
Why not simply pin the versions in `requirements.txt`?

danielmiessler:
Thank you, we'll check out the PR!

PR #32, opened 2024-02-04
Dependency management using Python Poetry for `fabric` cli

agu3rra:
## What this PR does
- dependencies for the `fabric` cli are now managed via [poetry](https://python-poetry.org/).
- `fabric` can be called directly as a CLI from the activated python virtualenv without the need to call it as `python fabric.py`.
- updates Fabric CLI install instructions.
- partially solves #27.

## Why so big?
- `Poetry.lock` maintains the hashes of all actually resolved dependencies. It's generated when one installs dependencies with `poetry add foo` automatically.

## What's in it for this project
- Deterministic dependencies pull and reproducibility: Since dependencies are resolved, locked and committed to source control, collaborators of this project can work on the same exact Python environment regardless of which OS they use and when they pulled dependencies.
- Isolated packages: Poetry helps avoid version conflicts in Python files and projects by isolating packages from the OS.
- Single configuration file: Poetry uses a `pyproject.toml` file to manage all dependencies, replacing multiple files like `setup.py`, `requirements.txt`, `setup.cfg`, and `Pipfile`.
- Accessible project metadata: Project metadata is easy to access and understand from the `pyproject.toml` file.

## What will the install actually look like
Screenshot 2024-02-04 081542

## What's next:
Do this for the Flask server app if accepted.


danielmiessler:
I think we might have addressed a number of these already, and we're getting conflicts on this one now. Can you look again and resubmit?

Issue #47, opened 2024-02-06
Typo in `README.md`


cleong14:
On #L119, `... systemClone ...` should be `... system. Clone ...`.

https://github.com/danielmiessler/fabric/blob/cf928e631fd0d9580eaf59015512d24dd47d802c/README.md?plain=1#L119

danielmiessler:
Thank you!

PR #48, opened 2024-02-06
fix: README typo; add extra files to .gitignore

cleong14:
fixes: #47
Signed-off-by: Chaz Leong <13462818+cleong14@users.noreply.github.com>


danielmiessler:
I think we already addressed this one?

Issue #44, opened 2024-02-06
api server only support extract_wisdom pattern? I think the api path should be dynamic


dheerapat:
we can do something like

```python
path_mappings = {
    "path1": {"system_url": "https://example.com/system1.md", "user_url": "https://example.com/user1.md"},
    "path2": {"system_url": "https://example.com/system2.md", "user_url": "https://example.com/user2.md"},
    # also this can be a relative path since the back-end server lives along pattern directory
}

@app.route("/<path_segment>", methods=["POST"])
def mill(path_segment):
```

just a rough idea, I will tackle this on the weekend. I want to contribute to this and improve quality of the back-end server.

danielmiessler:
Interesting idea, thank you!

PR #52, opened 2024-02-07
feature: make path-pattern mapping dynamic and scalable

dheerapat:
Currently your back-end API have only one, static path for external_wisdom pattern which is too rigid and hard to scale. As mentioned in #44 

I added a new way to dynamically map each path to each pattern and easy to scale further.

This is not a breaking change, It still compatible with the original code, you will be able to run the same HTTP call without error. But now you will have more flexibility on the back-end side.

### DIFF
--- a/server/openai.key
+++ b/server/openai.key
@@ -1,1 +1,1 @@
-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
+sk-somethingsomethingnumbersandstuff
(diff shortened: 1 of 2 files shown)


server/openai.key
dheerapat, line 1: I accidentally push this to the remote, already revoke

danielmiessler:
This has conflicts, unfortunately, and I'm not sure I'm fully grasping the idea. Can you restate?

dheerapat:
> This has conflicts, unfortunately, and I'm not sure I'm fully grasping the idea. Can you restate?

example: currently when I go to `localhost:5000/extwis` I can send a post request to tell the server to mill my input with your pattern and then send to OpanAI to get the answer

I proposed that instead you building `@app.route("/extwis", methods=["POST"])` for every pattern, we can just make path `/<pattern>` dynamic by adding a mapping dictionary to map path with the pattern user want to use. [more info on this](https://flask.palletsprojects.com/en/3.0.x/quickstart/#variable-rules)

PR #54, opened 2024-02-07
Update system.md

Mgmoser:
I added text to the #3 output section. Now the AI should append the person who said the quote to the end of the captured quote. I used the text "Incorporate the name of the person quoted at the conclusion of the quotation."


danielmiessler:
We usually don't have speaker information, actually, so we won't be able to do this quite yet. But please submit later!

Issue #67, opened 2024-02-10
Missing requirements.txt file for client install


natv:
In the README quickstart for the client install, it instructs to

# Enter the project and its /client folder
cd fabric/client

# Install the pre-requisites
pip3 install -r requirements.txt

However, there is no requirements.txt in the fabric/client folder. I did find one in the 'server' folder and installed those requirements, but perhaps something is still missing because after adding the alias (or even trying manually), it's not running, it complains that I'm referencing a folder when I do 'fabric -h'

$ fabric -h
-bash: /home/user/ai/fabric/client/fabric: Is a directory

Thanks

autumn-traveller:
Same issue here. Opened a PR to try to fix. https://github.com/danielmiessler/fabric/pull/69

agu3rra:
Hello folks!
The client now uses python poetry to manage dependencies instead of regular `requirements.txt`. Install instructions are on the client readme file [here](https://github.com/danielmiessler/fabric/blob/main/client/README.md).

I think this is just a matter of updating the main README file's instructions for the client.

danielmiessler:
#Can you submit a PR to update that readme?

agu3rra:
> #Can you submit a PR to update that readme?

@danielmiessler It's PR #72.

BryanKerrEdTech:
Does the new install instructions mean we no longer need to do the current Readme's Steps 5 and 6? Seems like the poetry shell makes those steps unnecessary, correct?

agu3rra:
@BryanKerrEdTech
Steps 5 and 6 seem to point to how to add the CLI to your PATH. If you're running the cli after running `poetry shell`, the virtualenv it creates will already know `fabric` and you're correct (no further action required).

However...
If you wish that `fabric` is a known cli command after you reboot your computer and you don't wish to activate the virtualenv every time, I suggest you add it to your path by running:
```bash
# Determine the path of your fabric CLI for the next step
# It should be something like /home/thanos/.cache/pypoetry/virtualenvs/fabric--_h0wFFW-py3.10/bin/fabric
which fabric
```

Then use the detected path to run steps 5 and 6.

BryanKerrEdTech:
Thanks so much for clearing that up for me Andre!

PR #69, opened 2024-02-10
add requirements.txt (from release 1.0.0) back to the client directory

autumn-traveller:
I noticed that the requirements.txt file was missing in the `client` folder, and saw that an issue had been opened about this as well (https://github.com/danielmiessler/fabric/issues/67). I have pushed a fix where I just copied back the `requirements.txt`  file from the v1.0.0 release into the `client folder`

### DIFF
--- a/client/requirements.txt
+++ b/client/requirements.txt
@@ -1,0 +1,17 @@
+pyyaml
+requests
+pyperclip
+python-socketio
+websocket-client
+flask
+flask_sqlalchemy
+flask_login
+flask_jwt_extended
+python-dotenv
+openai
+flask-socketio
+flask-sock
+gunicorn
+gevent
+httpx
+tqdm


agu3rra:
If this item is merged then #72 should not be and PR #65 should be reverted.
Alternatively, simple merge #72 and close this PR.

It's a philosophical thing. Either we manage dependencies with Poetry or via `requirements.txt` :)

autumn-traveller:
I agree, it should be one or the other. Can't say which would be better, but if we move to `poetry` then it would be good to update the `server` directory so that it also uses `poetry` and not a `requirements.txt`

agu3rra:
> I agree, it should be one or the other. Can't say which would be better, but if we move to `poetry` then it would be good to update the `server` directory so that it also uses `poetry` and not a `requirements.txt`

That's the next PR I plan on submitting. Cheers!

autumn-traveller:
Great stuff, then I'll close this PR and we move to `poetry` everywhere. And don't forget to update the installation instructions for the server in the Readme

PR #82, opened 2024-02-11
Update fabric.py to support YouTube Transcripts

DuckPaddle:
Added cli option --youtube -y that invokes the transcript function in utils.py with the video id specified on the command line


danielmiessler:
This caused breaking changes, so we had to remove.

DuckPaddle:
Can you provide some more detail on the testing?

VP Engineering

Hexicurity

***@***.***

[832-752-3084](tel:8323804878)

PR #88, opened 2024-02-12
Multiple changes: single poetry project; Bash script for creating `aliases`; updated instructions

agu3rra:
## What this PR does
- We now use a single Python Poetry project to manage all 3 python specific command line applications: `fabric`, `fabric-api`, and `fabric-webui`.
- Creates `setup-aliases.sh` to facilitate the addition of such commands as aliases in Unix-based systems. The goal is that users can run the commands without being in the active Python virtual environment created by Poetry.
- Updates main `README.md` install instructions.
- Redirects CLI install specific instructions to the corresponding section on the main `README.md` file so we don't have to maintain them in both locations.

## Screens
Running the commands outside of the poetry environment.

Screenshot 2024-02-11 213618

---

Running the alias creation bash script multiple times.

Screenshot 2024-02-11 221446


danielmiessler:
This is a massive change to the structure. Let's talk with @xssdoctor and discuss. For one thing, we lost the client directory, which I'd like to still have in the structure. We also appeared to lose the /patterns directory? I'm confused by the changes, and it says this changes 41 different files.

Let's have a call to discuss this one, and possibly making this multiple PRs with discrete changes rather than something this big. Thoughts?

chrissonntag:
Agreed, this is a major breaking change. Although the idea behind it is delighting. We can keep the restructuring in mind for a next major version, or a minor version with migration path.

Wouldn't this be an appropriate situation to establish a coding guideline incl. Issue-/PR-Templates and Contribution-Guideline?

agu3rra:
Hello folks!
I agree, we can make have a call to discuss. A few points from me ahead of that:
1. I re-added the client folder to the structure on my latest commit. `cli` and `gui` are in there.
2. The reason the PR has too many files changes is because I've moved them around, but it's "only" 212+ lines of code.
3. I am wondering how I'd make it smaller considering the idea here was to remove the 2 separate python poetry projects (1 for client, 1 for server) to make the experience less of a hassle (activating more than 1 `virtualenv` for the commands to be found).
4. I am not sure I get what the breaking change is. I mean, sure we now have a different install process, but simpler. Considering that a breaking change would also constitute the same type of issue when moving away from `requirements.txt` into poetry.

We can have a call and screenshare and I'll walk around what I thought. Perhaps even find a proper home to the `gui` npm-based folder.

Cheers!

chrissonntag:
Regarding #4: People are already struggling with getting the fabric client to run. If we were to restructure the directories, the entries in .bashrc/.zshrc wouldn't work anymore. This is what I've meant when talking about a breaking change. A pull would literally break the client for those not knowing what is going on.

agu3rra:
> Regarding #4: People are already struggling with getting the fabric client to run. If we were to restructure the directories, the entries in .bashrc/.zshrc wouldn't work anymore. This is what I've meant when talking about a breaking change. A pull would literally break the client for those not knowing what is going on.

Understood. Well, let me know if you wish to sync today. It's Carnival time here in Brazil, but I usually like shaking my fingers on the keyboard rather than my rear on the streets.

chrissonntag:
I'd be in for it as well, but it's up to Daniel. I'm in GMT+1 btw.

agu3rra:
@danielmiessler 
The latest commits renamed the `fabric` folder into `fabric_installer` as we discussed. It also updates the readme instructions accordingly. Please take a look and see if the instructions look clearer and if it works on your side as well (ideally in a brand new environment).

agu3rra:
@danielmiessler 
I tested the new install instructions on README.md inside a brand new Debian box and a not so new macOS running the latest Sonoma version. Please verify if it works on your end.

PR #102, opened 2024-02-14
turns module import into package import

### LINKED ISSUE #99: Can't find module 'utils' (Windows)[Solution]
## Environment
- **Operating System**: Windows
- **Python Version**: [3.11.5]
- **Poetry Version**: [1.7.1]
- **Fabric Version**: [0.1.0]

## Issue Description
When attempting to execute a sequence of commands involving Poetry and Fabric, an error is encountered stating that the module 'utils' cannot be found. This issue arises after initiating a Poetry environment and executing the `fabric` command.

### Steps to Reproduce
1. Run `poetry install` to install dependencies.
2. Activate the virtual environment with `poetry shell`.
3. Execute the `fabric` command.

#### Expected Behavior
The `fabric` command should execute without any issues, with all modules being correctly resolved.

#### Actual Behavior
An error message is displayed: `can't find module 'utils'`. This prevents the `fabric` command from executing as expected.

## Solution
The issue appears to be related to an incorrect import statement within the Fabric library. To resolve this issue, a modification in the import path within a specific file is required.

### Detailed Steps to Resolve
1. Navigate to the file located at `\fabric\client\fabric\fabric.py`.
2. In the first line of the file, locate the import statement for `utils`.
3. Modify the import statement from `import utils` to `import fabric.utils`.

After making this change, the `fabric` command executes successfully without encountering the module resolution error.

## Additional Notes
- This solution is specific to the described environment and versions. If the issue persists, please check for any updates or patches for the involved software.
- It's recommended to document any changes made to library files for future reference or in case of updates.

agu3rra:
I thought I had turned this into a package relative import when I introduced the poetry PR, perhaps it got lost in one of the merges?

Fixes and closes #99 

From `git blame` I see it was turned into an absolute import [on this commit](https://github.com/danielmiessler/fabric/commit/4c09fa3769140496d785db45f16a2a1dfe2ac32d). I wouldn't agree to it though. Happy to discuss further if necessary.

### DIFF
--- a/client/fabric/fabric.py
+++ b/client/fabric/fabric.py
@@ -1,4 +1,4 @@
-from utils import Standalone, Update, Setup
+from .utils import Standalone, Update, Setup
 import argparse
 import sys
 import os


agu3rra:
> Its not that simple, since adding fabric before utils, makes it work on windows but not MacOS.

@chroakPRO 
Can you please elaborate why you think that relative import doesn't work on macOS? It is an import that happens within the fabric package, so I don't understand what you mean when fabric is being added before utils.

I just cloned my fork and branch for this  PR on macOS (Sonoma 14.3.1) and the CLI works on the virtualenv just as intended.

Thank you.

chroakPRO:
Oh okey, maybe it works then on my system it didnt work.

chroakPRO:
Nevermind actually, the problem was that I had fabric as an alias with calls the fabric.py directly, thats why that change didnt work. You are correct this would work on mac aswell.

agu3rra:
Not required after we merge #88

PR #110, opened 2024-02-17
[FEATURE] Introduce LiteLLM to Fabric

chrissonntag:
Introduce LiteLLM to Fabric
- Use all by LiteLLM supported LLM Provider and the model of your choosing.
- Use OpenAI like before - No further setup needed.

Proposal of next steps:
- Implement an "Exception Handling" for the CLI - Pretty print error messages that are meaningful to the user.
- Extend utils to an entity that holds most of the environment variables that are needed to use the respective LLM Provider - I.e. define that huggingface needs HUGGINGFACE_API_KEY, etc.
- This entity and the env vars can then be used to build some kind of menu for the user to decide what env vars to define, what provider to use.
- The method fetch_available_models() needs a redesign. We should check which LLM Provider can be used and either get all models along all providers or let the user differentiate.

What has to be discussed:
- What are we going to implement and how do we do it?
- What coding standard are we going to follow?
- Are we planning of using github actions?

For now, I leave it there. I've seen in the README that the Backend is supposed to have a redesign and an implementation with FastAPI. Any ETO on that? I'd rather wait with the LiteLLM integration into the backend until that's done.

@agu3rra I am not that familiar with poetry. Therefore I am a little bit confused why the poetry.lock changed that much. Maybe you can enlighten me and fix the lock file if it even needs fixing.
@xssdoctor I'm living in GMT+1. Shall I contact you via Email to arrange a meeting with you and Andre?


agu3rra:
@chrissonntag 
`poetry.lock` maintains the hashes and all actually resolved libraries. It's what ensures every `poetry install` produces the exact same Python environment in different machines (including yours and mine) so we can work with the same environment. Can you confirm the new lib you added `litellm` was via `poetry add litellm`? The command would pull it and all related subdependencies, which is why `poetry.lock` would automatically change and be bigger.

README.md
```diff
@@ -422,6 +423,30 @@ The content features a conversation between two individuals discussing various t
...
 ```
 
+## LLM Providers / LocalLLMs
+`fabric` leverages LiteLLM to enable users to seamlessly utilize a wide array of LLM Providers and Models. With LiteLLM, users can effortlessly switch between different providers and models by specifying the desired provider and model name through command line options. `fabric` utilizes LiteLLM to dynamically load the specified provider and model, empowering users to harness the unique capabilities offered by each LLM provider or model.
+
+> Please ensure that you have configured the appropriate environment variables. For instance, set `HUGGINGFACE_API_KEY` if you intend to utilize the Hugging Face API.
+
+### Usage:
+
+To specify the provider and model when running `fabric`, use the `--model` option followed by the provider and model name.
```
agu3rra, line 433: How does a user list available models to select? Perhaps I need to read on LiteLLM, but I'd expect there'd be a `--list` option of some sort for models.

pyproject.toml
```diff
@@ -13,6 +13,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.10"
+litellm = "^1.24.6"
```
agu3rra, line 16: If litellm is to be used in the `cli` code only, please add it to the corresponding CLI group instead of the generic bucket of dependencies. Please do not edit this section directly with your text editor, but do the following instead (from the project's root folder):
```bash
poetry remove litellm
poetry add --group=cli litellm
```

installer/client/cli/utils.py
```diff
@@ -1,6 +1,6 @@
 import requests
 import os
-from openai import OpenAI
+import litellm
```
agu3rra, line 3: So the proposal is to switch from the `openai` package for the `cli` alltogether? If so, please remove it from the `cli` group dependencies by running:
```bash
poetry remove --group=cli openai
```

README.md
agu3rra: Never mind, I noticed there's a `--listmodels` option. Considering this closed.

agu3rra, line 433:
```suggestion
To specify the provider and model when running `fabric`, use the `--model` option followed by the provider and model name.
To list models, use the `--listmodels` option.
```

agu3rra:
From a project's perspective, I think it'd be interesting to use [Python click](https://click.palletsprojects.com/en/8.1.x/) as the framework for the CLI. As the functionality of the `fabric` cli grows, I think it can be easier to organize different functionalities in different subcommands. I'll get better acquainted with all the options and propose something later.

README.md
chrissonntag: Right, problem is... `--listmodels` only lists the available OpenAI models at this point. We have to implement some kind of helper functionality that differentiate between the configured providers.

pyproject.toml
chrissonntag: The plan is to use LiteLLM also for the server implementation. Wouldn't it be smarter to just leave LiteLLM than as global dependency?

installer/client/cli/utils.py
chrissonntag: Wouldn't we rather leave LiteLLM just right where it is now and remove openai from both groups? Where is the advantage of defining openai separately in the first place? I guess we could have different implemenations in the cli and the server. But are we really going to?

chrissonntag:
> @chrissonntag `poetry.lock` maintains the hashes and all actually resolved libraries. It's what ensures every `poetry install` produces the exact same Python environment in different machines (including yours and mine) so we can work with the same environment. Can you confirm the new lib you added `litellm` was via `poetry add litellm`? The command would pull it and all related subdependencies, which is why `poetry.lock` would automatically change and be bigger. It is normal.

I'm aware of the purpose if the poetry.lock, but am not familiar with the structure of it. So I guess everything is just fine and the fact that the poetry.lock doubled in size is due to the dependencies installed by LiteLLM.
And yeah, I can confirm that I added it via `poetry add litellm`.

installer/client/cli/utils.py
agu3rra: I agree. I'll submit a PR to unify all dependencies later then. It should keep things simpler now that we have a single poetry project for everything.

README.md
agu3rra: I'd understand this is a required item to ship this new feature. Otherwise we need to at least let people know which models are available for them to select since these are expected to be provided as text. Is the list the one provided at https://docs.litellm.ai/docs/providers?

Unless of course we make the openai one the default and people then don't need to manually select a provider. Then I think it'd be ok-ish with pointers to these docs or a smart way to retrieve the list of providers from the `litellm` lib itself. Perhaps we can invoke a method from it and make the list available.

chrissonntag: The link is correct. In the beginning we can lead people to that list. In the long run I'd prefer having an internal notification for that. 
As far as I can see, there's no proprietary way to do so. LiteLLM has its internal variables of providers but cannot deliver a list of models. What models are available to which provider depends on many factors. Such as Key Restriction or simply what Ollama models were locally downloaded.
So I guess we can try to enhance the user experience as best as we can. But we won't make it right for everyone  😅

installer/client/cli/utils.py
chrissonntag: Great 😁 Consolidation is always welcome 😅

danielmiessler:
So please walk me through how to use this in practice with a local model.

Where does the local model have to be listening? On what IP and ports? And how do you specify that when running fabric using your litellm implementation?

th3f001:
If we chose to use LM studio as local LLM multiplexer, I see it has its own server part accepting API calls.
LM_Studio_local_server

chrissonntag:
> So please walk me through how to use this in practice with a local model.
> 
> Where does the local model have to be listening? On what IP and ports? And how do you specify that when running fabric using your litellm implementation?

IMO Ollama is the easiest and most convenient way to run LLMs locally. You would have to install Ollama on your system, pull the model of your choosing and then run fabric with it.

1. Install [Ollama](https://ollama.com/)

2. Pull a model
Available models can be found here: https://ollama.com/library

```shell
ollama pull mistral
```

3. Run fabric with it
```shell
fabric --model ollama/mistral ...
```

There are a bunch of providers like Azure and Huggingface where you would not need to have a model on your local system, where you would run the inference via an API just like OpenAI API. That's the big advantage of having LiteLLM integrated.
And although LiteLLM supports any model served with a [OpenAI-Compatible Endpoint](https://litellm.vercel.app/docs/providers/openai_compatible), there would not be a convenient way to allow users to run their models with tools like [LM Studio](https://lmstudio.ai/) or [Text generation web UI](https://github.com/oobabooga/text-generation-webui). Thus I think we should leave it to the user to either use all those already supported [providers](https://litellm.vercel.app/docs/providers) or - for advanced users - to integrate their endpoints by themselves. Maybe we can implement a hook to do so?!

chrissonntag:
I have integrated the last changes made on Feb 21, so that the branch does not have anymore conflicts. Please consider merging this PR before making further changes.

PR #132, opened 2024-02-23
make location of config files configurable

hv10:
Allows user to configure location of config directory.

## Why:
I wanted to have multiple "projects" going on - each of which with their own configuration (API endpoint / available patterns). To facilitate this I added the `--cfg-dir` optional argument.

Without this argument the default stays at `~/.config/fabric`.


hv10:
Addendum: I made the path construction like in the original code, as otherwise it is probably not system independent.

PR #150, opened 2024-03-03
Verify that poetry was setup correctly

breadtk:
Add basic sanity check to see whether or not `poetry` is setup correctly.

## Related issues
n/a

## Screenshots
n/a

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Verify that poetry is installed correctly.
+poetry --version >/dev/null 2>&1
+if [ $? -ne 0 ]; then
+    echo 'Poetry may not be installed correctly. Verify that poetry is in your $PATH.'
+    exit 1
+fi
+
 # Check if pyproject.toml exists in the current directory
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Verify that poetry is installed correctly.
+poetry --version >/dev/null 2>&1
+if [ $? -ne 0 ]; then
+    echo 'Poetry may not be installed correctly. Verify that poetry is in your $PATH.'
```
verebes1, line 6: Perhaps there could be some extra troubleshooting output like:
Check your .zshrc or .bashrc contains export PATH="$HOME/.local/bin:$PATH"
If it is there try `source ~/.zshrc` or `source ~/.bashrc`
Check if you have python3 installed by running `python3 --version` and try installing poetry again using:
curl -sSL https://install.python-poetry.org | python3 -

Merely a suggestion :)

danielmiessler:
We just moved to pipx, can. you see if we still need this and resubmit if we do?

Issue #155, opened 2024-03-04
[Feature request]: allow the default model to be set with env variables


xntrik:
### What do you need?

allow the default model to be set with env variables

xssdoctor:
Done

xssdoctor:
Added as —changeDefaultModel. All the available models are listed in —listmodels. It’s also added to —setup

PR #156, opened 2024-03-04
feat: allow setting a default FABRIC_MODEL via env vars (or .env)

xntrik:
Allow the default model to be set with env variables

## Related issues
#155


xssdoctor:
done

PR #165, opened 2024-03-07
feat: install poetry when running setup.sh

SamyGhannad:
This PR adds an `if` block to `setup.sh`, it checks if Poetry is already installed, if not, it asks if the user wants the script to install it for them.


danielmiessler:
We've moved to pipx so I think his should be solved

PR #189, opened 2024-03-12
Add code to use openai_base_url and use OpenAI's model lister function

zestysoft:
I had some trouble running fabric locally due to some hard coded openai urls in the fetch_available_models functions which required having a valid OPENAI_API_KEY.

This should let fabric communicate with other open_api compliant servers.

I also threw in replacement code to use the model listing function that's included with the OpenAI class.

Finally I've added some detail to the error messages if an exception is thrown while retrieving the list of models.

### DIFF
--- a/installer/client/cli/utils.py
+++ b/installer/client/cli/utils.py
@@ -1,6 +1,6 @@
 import requests
 import os
-from openai import OpenAI
+from openai import OpenAI, APIConnectionError
 import asyncio
 import pyperclip
 import sys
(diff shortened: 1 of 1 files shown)


meirm:
This commit generates an error:
fabric --setup                                                             
Traceback (most recent call last):
  File "/Users/meirm/Library/Caches/pypoetry/virtualenvs/fabric-M01m3Wmw-py3.11/bin/fabric", line 3, in <module>
    from installer import cli
  File "/Users/meirm/git/github/fabric/installer/__init__.py", line 1, in <module>
    from .client.cli import main as cli
  File "/Users/meirm/git/github/fabric/installer/client/cli/__init__.py", line 1, in <module>
    from .fabric import main
  File "/Users/meirm/git/github/fabric/installer/client/cli/fabric.py", line 1, in <module>
    from .utils import Standalone, Update, Setup, Alias
  File "/Users/meirm/git/github/fabric/installer/client/cli/utils.py", line 273
    print(f'{e.message} trying to access {e.request.url}: {getattr(e.__cause__, 'args', [''])}')

meirm:
This was the line:
                print(f'{e.message} trying to access {e.request.url}: {getattr(e.__cause__, 'args', [''])}')

And the fix:
                print(f"{e.message} trying to access {e.request.url}: {getattr(e.__cause__, 'args', [''])}”)

And there was another instance but the code changed and now with the new merges it is working again

zestysoft:
Looks like this PR got clobbered here
https://github.com/danielmiessler/fabric/commit/73c505cad12789a5b5dba275f3ba7d00664c8a19

I'll re-submit and apply that fix, ty!

PR #199, opened 2024-03-12
Add code to use openai_base_url and use OpenAI's model lister function

zestysoft:
Looks like my previous pr got clobbered https://github.com/danielmiessler/fabric/pull/189#issuecomment-1992514728


danielmiessler:
Really like this, but want to make sure @xssdoctor approves it as well and that we have documentation for it and evereything.

xssdoctor:
Love it. your way is much better

Issue #198, opened 2024-03-12
[Bug]: sed:  in-place editing only works for regular files


zestysoft:
### What happened?

./setup.sh assumes that the dot files to be edited are regular files when they could be symlinked.

### Version check

- [X] Yes I was.

### Relevant log output

```shell
sed: /Users/username/.zshrc: in-place editing only works for regular files
sed: /Users/username/.zshrc: in-place editing only works for regular files
```

meirm:
I found the bug and will  send a pull request soon

danielmiessler:
Please resubmit PR

PR #202, opened 2024-03-13
Fix bug on setup.py when editing symlinks

meirm:
Resolve symlink before calling sed
## Related issues
If it closes an issue, type `closes #[ISSUE_NUMBER]`.](sed: in-place editing only works for regular files #198)

## Screenshots
Provide any screenshots you may find relevant to facilitate us understanding your PR.


danielmiessler:
Not using setup.py anymore. Now using pipx. Should fix your issue.

Issue #197, opened 2024-03-12
[Bug]: --copy parameter throwing an error


arcaneum:
### What happened?

when I try to use the --copy parameter to pipe the output to the clipboard, I am getting the following error. I am running this on an Apple Silicon Mac with M3 Max chip.

The exact command I used on the Mac terminal was:

> yt 'https://www.youtube.com/watch?v=lEXd6TXPw7E' | extract_wisdom --copy

### Error Message
```
Traceback (most recent call last):
  File "/Users/<username>/Library/Caches/pypoetry/virtualenvs/fabric-xqnZk2pv-py3.11/bin/fabric", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/<username>/fabric/installer/client/cli/fabric.py", line 146, in main
    standalone.sendMessage(text)
  File "/Users/<username>/fabric/installer/client/cli/utils.py", line 246, in sendMessage
    pyperclip.copy(response.choices[0].message.content)
                   ^^^^^^^^
UnboundLocalError: cannot access local variable 'response' where it is not associated with a value
```

### Version check

- [X] Yes I was.

meirm:
The code needs some cleanup. 
In the meantime I posted a pull request https://github.com/danielmiessler/fabric/pull/203

xssdoctor:
OK, i added --copy for local models and for claude

arcaneum:
Thank you, @xssdoctor

PR #203, opened 2024-03-13
Fix bug in sendMessage by moving code

meirm:
Move code related to openai inside the try block
Note: the current implementation of the code allows the use of copy and output only when using openai

## Related issues
[Bug]: --copy parameter throwing an error #197

## Screenshots
NA


danielmiessler:
I like it, let's confirm with @xssdoctor

Issue #178, opened 2024-03-11
[Feature request]:  LocalLLM(ollama) support


rony432:
### What do you need?

this is awesome project, but it needs ollama support.
the OpenAI api is the ease way out.

please add support for local LLMs too.

thank you

xssdoctor:
It has that support. Ollama and Claude. —model. For a list of models run —listmodels. If you have ollama running on your machine it will list the ollama models you have running.

rony432:
i have ollama but not on this machine(its remote). i am looking to configure it as custom endpoint.
is there a sample config?

xssdoctor:
I don’t have a custom endpoint coded in yet. I was looking through the ollama python library and I couldn’t find a way to do it. If anyone knows how to do this I’ll add it

rony432:
check this out
https://docs.crewai.com/how-to/LLM-Connections/#setting-up-ollama

side quest. integration as tools for crewai to use fabric

gcapnias:
I believe ollama support needs some refining:

- You cannot use `--listmodels` unless you provide an OpenAI key. There are various sources of models; I get that. If there is not a key for a source, skip it, provide a message - there is no key for the group - continue to the next source.
- The model names returned from ollama are not correct; they are missing the last letter of their names. For some reason only llama2 returns the correct name.

Some enhancements:
- I would like to see the default model name reported from fabric if there is one. If not, return there is no default model.
- There is no command to get the version of fabric; Version is needed to report where the bugs are reported.

I used ollama v0.1.28, latest commit of fabric. Ollama has multiple models pulled:
```
codellama:latest        8fdf8f752f6e    3.8 GB  4 weeks ago
gemma:latest            430ed3535049    5.2 GB  2 weeks ago
llama2:latest           78e26419b446    3.8 GB  4 weeks ago
mistral:latest          61e88e884507    4.1 GB  6 weeks ago
neural-chat:latest      89fa737d3b85    4.1 GB  6 weeks ago
phi:latest              e2fd6321a5fe    1.6 GB  4 weeks ago
qwen:latest             d53d04290064    2.3 GB  4 weeks ago
```
Fabric reports these models:
```
Local Models:
codellam
gemm
llama2
mistr
neural-ch
phi
qwen
```

Looking forward for the next drop,
George J.

xssdoctor:
fixed. try it now without an api key and let me know if it works

ghzgod:
[...]

PR #225, opened 2024-03-14
Allow an empty OpenAI API key

PatrickRuddiman:
# Changes Made in this Commit

This commit contains changes to the `utils.py` file located in the `installer/client/cli` directory. The main modifications include:

1. Added error handling for cases where the API key is not set, preventing connection to the OpenAI API server.
2. Updated the model selection process to only retrieve models if an API key is present. If no API key is found, an empty list is returned instead.

# Reason for Changes

These changes were made to improve error handling and provide more informative messages when attempting to connect to the OpenAI API server without an API key set. This should help prevent issues running only with local models.

# Impact of Changes

These changes may improve the user experience by providing clearer error messages when connecting to the OpenAI API server without a valid API key. They also ensure that model selection is only performed if a valid API key is present, reducing potential errors and improving overall application functionality.

# Test Plan

These changes were tested locally by attempting to run `fabric` without setting an API key and observing the error messages produced. Additionally, tests were carried out with a valid API key set to ensure that model selection worked correctly. 

This would solve the 401 errors called out in #178 when listing the models

### DIFF
--- a/installer/client/cli/utils.py
+++ b/installer/client/cli/utils.py
@@ -259,6 +259,8 @@ class Standalone:
                     with open(self.args.output, "w") as f:
                         f.write(response.choices[0].message.content)
         except Exception as e:
+            if getattr(e.__cause__, 'args', [''])[0] == "Illegal header value b'Bearer '":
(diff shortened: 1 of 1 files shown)


danielmiessler:
This has been added already. Thank you!

PR #257, opened 2024-03-16
Webscraper

bthrx:
Created a helper for scrapping websites using beautiful soup to pipe into fabric.

### DIFF
--- a/helper_files/README.md
+++ b/helper_files/README.md
@@ -89,3 +89,26 @@ tags: fabric-extraction stub-for-name extra-tag
 ---
 test
 ```
+
+## ws (Web Scraper)
+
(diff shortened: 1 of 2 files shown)


xssdoctor:
Looks great! I also made some changes to add it to the poetry config so when you pipx install it adds ws to the path

bthrx:
Looks like it ended up on the dependabot branch.

joaomorossini:
I was searching for EXACTLY this feature! I've updated fabric and tried to use the ```ws``` helper app, but I just get ```command not found```.

**How do I install it? Can anyone give me a hand?**

I'm very excited to starting using it with Fabric! In fact, in my opinion, this is the ONE big feature Fabric was missing


Thanks in advance :)

PR #298, opened 2024-03-25
Fixed Latin-1 decode problems

Loibl33:
Fixes Latin-1 decode problems

Fixes Latin-1 decode problems

## Related issues
For example, https://youtu.be/f2BjFilLDqQ threw a Latin-1 decode error.

## Screenshots
Old behaviour: 
image

New behavior
image

### DIFF
--- a/installer/client/cli/yt.py
+++ b/installer/client/cli/yt.py
@@ -103,7 +103,7 @@ def main_function(url, options):
         if options.duration:
             print(duration_minutes)
         elif options.transcript:
-            print(transcript_text)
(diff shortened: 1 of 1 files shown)


CaeChao:
Hi @danielmiessler  @Loibl33,  I think this error is very likely caused by the ` locale` and/or environment is broken, not installed, not set to `UTF-8`, see [here](https://stackoverflow.com/questions/35502328/unicodeencodeerror-on-linux-but-not-on-windows/35508945#35508945)
Therefore, this fix might be unnecessary and will also break other languages output: see #370

PR #368, opened 2024-04-26
fix: The variable 'wisdomFilePath' is already a complete path constru…

RealHurrison:
The variable 'wisdomFilePath' is already a complete path constructed with 'config_directory'. Joining it again with 'current_directory' could lead to an incorrect path.


ksylvan (approved):
Good catch. Looks good to me.

PR #537, opened 2024-06-04
Add a pdf Extractor script to directly pipe pdfs into patterns

WielandF:
Add a pdf extractor script to directly pipe pdfs into patterns. This script is integrated into the pipx setup and can be called via "pe filename" in bash as shown in the readme. 

```bash
pe filename | fabric -p extract_wisdom 
```
~~Currently only basic extraction is functional~~
Increased functionallity  with markdown extraction from documents and saving of the markdown if wished (Flags: -m for markdown and -s to save created markdown. If no flag is set extraction happens without markdown).
 ```bash
pe filename | fabric -p extract_wisdom -m -s
```

## Screenshots
image


danielmiessler:
Moving to Go soon; let's hold off on this one. But Thank You for submitting. Let's revisit soon!

christrille:
Right now I am using extract_text function from pdfminer.high_level (see code bewlow).

I tried with PyPDF2 PdfReader function, but results were strange on occasion (one word per line).
Looking forward to see the "pe" function working.

CODE SAMPLE:
from pdfminer.high_level import extract_text
def pdf_to_clean_text(pdf_path):
    extracted_text = extract_text(pdf_path)
    return extracted_text

WielandF:
@christrille you can try out the "pe" function in my fork if you want and give feedback if you find strange behavoir.
As soon as @danielmiessler is ready to integrate it into the original fabric you can use "pe" here as well :)

PR #541, opened 2024-06-04
Adds pattern to generate pattern

Saik0s:
This pull request adds a new pattern with pattern template. It lets user create new patterns from unstructured instructions.


danielmiessler:
Please resolve

PR #559, opened 2024-06-06
feat: add create_pattern pattern

blade1981m:
This pull request includes a pattern to create new fabric patterns


blade1981m:
This was designed to be used with the output of the improve_prompt pattern though it does not need to be. Your stitch could be:

echo "Your custom prompt here" | improve_prompt | create_pattern

I have managed to make a couple of personalized patterns for myself using this pattern and had good results, but there is definitely room for improvement and so welcome any proposed changes.

PR #600, opened 2024-06-13
Update patterns\clean_text\system.md to improve text cleaning instructions

Gerkinfeltser:
This is a very simple pull to get my "pull request" feet wet! 😄 

This adds a single line to the `clean_text` pattern that instructs the llm add capitalization, paragraphs, and other formatting. This makes the output  much more readable and works well with youtube transcripts.

## Example:
(The following is using `meta-llama/llama-3-70b-instruct` as the model for both outputs.)
```
yt "https://www.youtube.com/watch?v=8f5DKSx_nEE" | fabric -p clean_text
```
### With the added line:
```
Here is the properly formatted text:

this is NASA's Boeing crew flight test mission, it the very first flight of astronauts aboard Starliner to the International Space Station. The past 24 hours have been a testament to what it's about in human space flight and really all the hard work that they've poured into this spacecraft over many years to get us to this point.

Toward the end of the day yesterday, uh, I would say it was around 4:30 local time here in Houston, after they had performed some of the demos. We noticed a little bit of increase in the helium leak rate that we talked pre-launch. Um, what we found also was two other leaks in Port manifold one, and that leak was a little larger than the ones we had seen before.

As a precaution, because we didn't need the capability to have all the thrusters, we're just maintaining attitude, pointing the tail toward the sun. Now, we recovered four of these five thrusters. So what we did is, once the Thruster fails ? when I say Thruster fails, what I mean is the software sees something that it doesn't like about that Thruster ? so the Thruster, uh, is maybe a little less thrust or the thrust rise rate doesn't come up exactly the way that the software is looking for.

I think we'll have to take a little bit more time to figure out what's going on with the thrusters. Uh, again, the rest of the Rendezvous went really well. The Vesta system, which is navigation system, worked great the whole time. Uh, all those sensors performed well. Um, the NDS system, the NASA docking system, which Jim talked about, is going to be used for Orion, worked extremely well. That system performed nominally. Um, and so most of the vehicle did awesome going into the Rendezvous.
```
### Without the line:
```
Here is the properly formatted text:

this is NASA's Boeing crew flight test mission, it the very first flight of astronauts aboard Starliner to the International Space Station. The past 24 hours have been a testament to what it's about in human space flight and really all the hard work that they've poured into this spacecraft over many years to get us to this point. Toward the end of the day yesterday, uh, I would say it was around 4:30 local time here in Houston, after they had performed some of the demos, we noticed a little bit of increase in the helium leak rate that we talked pre-launch. Um, what we found also was two other leaks in Port manifold one, and that leak was a little larger than the ones we had seen before. As a precaution, because we didn't need the capability to have all the thrusters, we're just maintaining attitude, pointing the tail toward the sun. Now, we recovered four of these five thrusters. So what we did is, once the Thruster fails, when I say Thruster fails, what I mean is the software sees something that it doesn't like about that Thruster, so the Thruster, uh, is maybe a little less thrust or the thrust rise rate doesn't come up exactly the way that the software is looking for. I think we'll have to take a little bit more time to figure out what's going on with the thrusters. Uh, again, the rest of the Rendevous went really well. The Vesta system, which is navigation system, worked great the whole time. Uh, all those sensors performed well. Um, the NDS system, the NASA docking system, which Jim talked about, is going to be used for Orion, worked extremely well. That system performed nominally. Um, and so most of the vehicle did awesome going into the Rond do.
```

### DIFF
--- a/patterns/clean_text/system.md
+++ b/patterns/clean_text/system.md
@@ -6,6 +6,7 @@ You are an expert at cleaning up broken and, malformatted, text, for example: li
 
 - Read the entire document and fully understand it.
 - Remove any strange line breaks that disrupt formatting.
+- Add captialization, punctuation, line breaks, paragraphs and other formatting where necessary.
 - Do NOT change any content or spelling whatsoever.
 
 # OUTPUT INSTRUCTIONS


silverstreak (approved):
Looks fine.

PR #602, opened 2024-06-14
Improved handling of non-ASCII characters in transcript printing and JSON output from `yt` command 

JFK:
## What this PR does
Improved handling of non-ASCII characters in transcript printing and JSON output from `yt` command 

- Updated print(transcript) to handle non-ASCII characters correctly.
- Modified print statements to use ensure_ascii=False in json.dumps for correct display of non-ASCII characters.
- Implemented exception handling to manage potential UnicodeEncodeErrors, ensuring fallback encoding to avoid garbled text.

This change enhances the readability of outputs containing non-ASCII characters and addresses potential encoding issues across different environments.

## Related issues and Screenshots
See this issue: https://github.com/JFK/fabric/issues/1

### DIFF
--- a/installer/client/cli/yt.py
+++ b/installer/client/cli/yt.py
@@ -110,11 +110,20 @@ def main_function(url, options):
         if options.duration:
             print(duration_minutes)
         elif options.transcript:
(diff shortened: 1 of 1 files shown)


danielmiessler:

PR #603, opened 2024-06-14
3 feature request support local openai whisper `ts` command

JFK:
Add support for local openai-whisper in ts command

- Updated `README.md` to include instructions for using local openai-whisper.
- Added error handling to `ts.py` for cases where `import whisper` fails.
- Modified `ts.py` to support the use of the local openai-whisper model.
- Added new command-line arguments to `ts.py` for specifying language and using the local openai-whisper model.

To use the local openai-whisper, run:
pipx inject fabric openai-whisper

This update allows users to choose between the accuracy of the OpenAI API and the cost-effectiveness of local processing.

## Related issues
https://github.com/JFK/fabric/issues/3


danielmiessler:

PR #604, opened 2024-06-14
Add option to specify output language for CLI input

JFK:
I am not sure if this is the best approach, but I wanted to propose it for consideration.

Add option to specify output language for CLI input

- Added `--output-language` (`-L`) argument to specify the output language for CLI input. Default is English.
- Updated `main` function to include the output language instruction in the input text if specified.

This allows users to set the desired output language for responses directly from the command line.

## Related issues
https://github.com/JFK/fabric/issues/5


silverstreak:
Would it not be possible to simply add a line in the system.md pattern file text indicating what the output language should be?

JFK:
@silverstreak 
> Would it not be possible to simply add a line in the system.md pattern file text indicating what the output language should be?

I tried to add a line into system.md but it did not work...
Anyway I can understand English.

silverstreak:
I'll also try it at my end and let you know. But in general, if it works, then it should be fairly easy to provide an input parameter like --output-language and then programmatically append/prepend "The output should be in xxx language" to the system.md file text that is sent to the model.

Gerkinfeltser:
> ... to the system.md file text that is sent to the model.

I've don't similarly with a context.md (& the `-c` flag) to good effect. I was getting some French responses prior but not when using my modified context file.

Side note: I saw someone mention adding functions to specify different context files. This seems like it would be super useful in the case of switching response languages per fabric command (among other things).

silverstreak:
> I'll also try it at my end and let you know.

OK so here's the result of my investigation.

1. --context option
   - The documentation on the --context option is quite sparse (almost non-existent)
   - Created context.md in ~/.config/fabric
   - Added this to it: # FINAL INSTRUCTION \n All the output should be in French.
   - Command used: pbpaste | fabric -c -p extract_wisdom
   - Outcome: nothing happens. The context file appears to be ignored
   - Note: maybe I'm using --context wrong
 2. Append a line to an existing pattern
   - Used extract_wisdom as an example
   - Appended the same line to the end of it: # FINAL INSTRUCTION \n All the output should be in French.
   - Outcome: output is in French

So I really think the most intuitive way to implement this feature (and actually, more generally, to add custom instructions to any pattern) is to do it this way: append whatever custom instructions you want to the system.md text that is sent to the model. Allow the user to input them via a file and using an option like --custom_instructions. My 2 cents.

danielmiessler:

Issue #468, opened 2024-05-30
[Feature request]: Multilingual yt - formatting issues


julliannailluj:
### What do you need?

I'm trying to use the yt --transcript function in languages other than English (French). Transcription contains formatting problems:
- writes "Ã©" instead of "é", "Ãª" instead of "ê", "Ã" instead of "à"...
- no punctuation

I tried to fix it with 2 different approaches:
- using a custom pattern to fix those typos.
- editing the yt.py by including a function to replace the defective set of typos with the correct ones.

From my attempts, the best fix for now is using a pattern. I tried it in 2 ways, natural language, and asking it to mimick a given python function. The second solution worked best. But it's not perfect and works in a very random manner. It often fixes the formatting problem, but sometimes randomly changes a small amount of words (with other words). Also, the puntuation is usually simply lacking, or not as good as it is in English. Those problems happens even when Youtube has a correct subtitle file in French. It also sometimes doesn't work at all and gives me comments regarding the python function.

The results are encouraging, but very random. I'm willing to improve it but maybe it's not the right approach. Any suggestions are welcome.

**Here is an example of a command I used:**

`yt --transcript --lang 'fr' https://www.youtube.com/watch?v=oiKj0Z_Xnjc | fabric --model llama3:latest -sp convert_fr`

**And the output:**

> `It seems you're asking me to run the fix_encoding_issues function on this text. I'll do that for you.
> 
> Please note that the original encoding of the text is not specified, but based on the presence of non-ASCII characters (e.g., Ã¢, Ã¨, Ãª, etc.), I assume it's encoded in UTF-8.
> 
> Here's the cleaned text:
> 
[...]

PR #614, opened 2024-06-17
Fix yt transcript encoding errors

fitchmultz:
Fix error in the encoded output when using the yt command with --transcript. 

Reproduce the error with the command 
`yt --transcript https://youtu.be/k8cdByDa3oA\?si\=__-Tlmvt7KoP7eR2`

Original Output: 
`In the early years of microelectronics,Â  there were many types of transistors. But one rose up to rule all the others. Based onÂ  an idea that physicists have chased for decades. And today, we have made more of it than anythingÂ`

Fixed Output:
`In the early years of microelectronics,  there were many types of transistors. But one rose up to rule all the others. Based on  an idea that physicists have chased for decades. And today, we have made more of it than anything `

Solution:
Edit `fabric/installer/client/cli/yt.py` line 113.
Original code: `print(transcript_text.encode('utf-8').decode('unicode-escape'))`
Fixed code: `print(transcript_text)`
Reinstall: `pipx install . --force`

Thanks to CaeChao here: https://github.com/danielmiessler/fabric/issues/468#issuecomment-2140722991

### DIFF
--- a/installer/client/cli/utils.py
+++ b/installer/client/cli/utils.py
@@ -407,7 +407,7 @@ class Standalone:
         fullOllamaList = []
         googleList = []
         if "CLAUDE_API_KEY" in os.environ:
-            claudeList = ['claude-3-5-sonnet-20240620','claude-3-opus-20240229', 'claude-3-sonnet-20240229',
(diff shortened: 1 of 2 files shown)


danielmiessler:

PR #615, opened 2024-06-17
Remember my remote Ollama server:port in .env

RossLux:
### RATIONALE
In order to streamline local LLM use on servers, here is a tiny update that simplifies calling remote Ollama models from the terminal machines. You can do so by setting the --remoteOllamaServer flag in each fabric call... every time you do a call. For me personally it's 40 extra characters that include the --remoteOllamaServer flag plus IP and port.
To further simplify this process, I propose adding support for the REMOTE_OLLAMA_SERVER environment variable, which enables users to configure their remote Ollama server settings globally. This enhancement fits well within the general quick setup paradigm and caters to a broader range of users who may not require this feature in every situation.

### CHANGES
- Added support for `REMOTE_OLLAMA_SERVER` environment variable.
- Updated `streamMessage` to use `REMOTE_OLLAMA_SERVER` if available.
- Refactored `Standalone` class to prioritize `REMOTE_OLLAMA_SERVER`.
- Simplified client initialization with `ollama.Client`.
- Ensured backward compatibility with existing `remoteOllamaServer` argument.

### DIFF
--- a/installer/client/cli/fabric.py
+++ b/installer/client/cli/fabric.py
@@ -178,31 +178,31 @@ def main():
         text = standalone.get_cli_input()
     if args.stream and not args.context:
         if os.environ["REMOTE_OLLAMA_SERVER"] or args.remoteOllamaServer:
-            standalone.streamMessage(text, host=os.environ["REMOTE_OLLAMA_SERVER"] or args.remoteOllamaServer)
+            standalone.streamMessage(text, host=args.remoteOllamaServer or os.environ["REMOTE_OLLAMA_SERVER"])
         else:
             standalone.streamMessage(text)
         sys.exit()
     if args.stream and args.context:
         with open(config_context, "r") as f:
             context = f.read()
-            if args.remoteOllamaServer:
+            if os.environ["REMOTE_OLLAMA_SERVER"] or args.remoteOllamaServer:
                 standalone.streamMessage(
-                    text, context=context, host=args.remoteOllamaServer)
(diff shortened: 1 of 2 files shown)


installer/client/cli/fabric.py
```diff
@@ -177,32 +177,32 @@ def main():
     else:
         text = standalone.get_cli_input()
     if args.stream and not args.context:
-        if args.remoteOllamaServer:
-            standalone.streamMessage(text, host=args.remoteOllamaServer)
+        if os.environ["REMOTE_OLLAMA_SERVER"] or args.remoteOllamaServer:
```
KasperZutterman: Calling `os.environ["REMOTE_OLLAMA_SERVER"]` will result in a `KeyError: 'REMOTE_OLLAMA_SERVER'` when a user doesn't have the `REMOTE_OLLAMA_SERVER` key in their `.env` file.

You should be using something like `os.environ.get('REMOTE_OLLAMA_SERVER', None)` instead.

(This comment is applicable for all similar calls you added.)

RossLux: Agree, thanks! I replaced the `os.environ["REMOTE_OLLAMA_SERVER"]` calls with `os.environ.get('REMOTE_OLLAMA_SERVER', None)` as you suggested.

Additionally, I added all of that to the PraisonAI agents section, replacing the `http://localhost:11434/v1` to make agents call the remote Ollama server if needed.

danielmiessler:

rcarmo:
Could we have this merged for those of us who have working Python installs with a private Llama server? It really does make a difference...

Issue #676, opened 2024-07-01
[Feature request]: A pattern that suggests what fabric pattern/command to use


kevnk:
### What do you need?

As a newbie to this, I basically only use `extract_wisdom`, but I know there's got to be more. 

I'd love to run something like `fabric suggest "I have a youtube video that I just want to get the main how-to steps out of it"`

I would imagine the output being either:
1) Here are some commands you could run `yt "YOUR VIDEO" | some_command` (maybe it shows and explains multiple options OR maybe it just says, "I got you, paste in the video URL and I'll run this command: `some_command`")
2) No command is found and so it runs `create_pattern` for you (bonus: if it prompts you to save this somewhere and copy it into the patterns directory. Double bonus: if it does that and then runs it!)

I'm thinking the `fabric suggest` command may dynamically cache some version of `create_summary` and/or `explain_docs` loop through all the patterns as context for suggesting the right command. 

Or maybe if you _really_ wanted to get fancy, you could set up a vector database for all the fabric commands to be able to search it better with real language without pinging your model. (This is new stuff to me and I'm not sure where I would start)

**BOTTOM LINE:** I want to use fabric more, but I don't want to spend time reading through all the system.md to guess what it's going to do.

kevnk:
Running this now...

`"I want a pattern that suggests what fabric pattern (https://github.com/danielmiessler/fabric/) I should use. The input would be what the user is looking for the pattern to do and the output would be a list of suggested fabric cli commands what you could run with an explaination of each. If there are no patterns that would seem relevant, then it could suggest `create_pattern` to create a new one and give instructions on storing/coping the pattern accordingly to get it working correctly. There should be a context.md file included in this pattern too that has a summary of all the docs and a summary of each of the commands" | fabric -sp create_pattern`

kevnk:
That wasn't all that great. I'll keep trying. For reference, here's the first version it came up with:

```markdown
# IDENTITY and PURPOSE

[...]

PR #678, opened 2024-07-01
Add 2 patterns: `summarize_prompt` which was used for `suggest_pattern` 

kevnk:
My goal is use a fabric command to learn how to better use fabric and all of its commands. 

1. I created a pattern to summarize the purpose, nuances, and output of a prompt
2. I attempted to create a 2nd pattern that would suggest a pattern when I wanted to do something (eg. Find the pattern that may help me list the how-to steps in a video) and if there were no patterns, then give suggestions for `create_pattern`
3. I created a script to use the `summarize_prompt` pattern to summarize each of fabric's patterns for use in the `suggest_pattern` prompt

## Related issues
#676


kevnk:
Thanks Daniel,

For the record, the `summarize_prompt` works well, but I don't think the `suggest_pattern` is honed in/working just right, yet. I was confused if/how to use the `user.md` file (which is where I put all the prompt summaries). I imagine if you can use the user.md file, it may work alright. 

Let me know what changes you suggest and I can make it happen. 

Thanks!

PR #687, opened 2024-07-03
Added current model indicator with an * when --listmodels is called

### LINKED ISSUE #446: [Feature request]: Asterisk next to default model in `fabric --listmodels` output
### What do you need?

Apologies if functionality is provided some other way, but it would be very user friendly and intuitive if an asterisk was printed next to the default model in the `--listmodels` output.

mchoccie:
Trying to add a small and simple feature for my first contribution. Whenever 'fabric --listmodels' is called, the output will denote the currently used model with an asterisk after it. Hopefully this will give users a better idea of the current model they've set Fabric to utilize.

## Related issues
closes #446 

## Screenshots
image

### DIFF
--- a/installer/client/cli/fabric.py
+++ b/installer/client/cli/fabric.py
@@ -159,32 +159,22 @@ def main():
             print("No patterns found")
             sys.exit()
     if args.listmodels:
+        def print_models(model_type, curr_model):
+            for model in model_type:
+                if curr_model == model:
+                    print(model + " (current)")
+                else:
+                    print(model)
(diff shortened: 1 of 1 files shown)


auryn31:
Just encountered the same thing and would love having it in. Would maybe write a model print helper function, to not have so many if repetitions and only one place to change the pattern of what is the current model and what not
maybe would also have instead of a * something like _(current)_

mchoccie:
Hey @auryn31 , I implemented the changes you mentioned. What do you think?

danielmiessler:

PR #709, opened 2024-07-09
Add feat: use Rich to enhance Markdown output

Tianyijian:
Markdown output is a suitable format, but it can sometimes be difficult to read in the terminal. 
Why not utilize [Rich](https://github.com/Textualize/rich) to enhance the markdown output and make it more readable and beautiful?

### changes:
- Add the Rich dependency in `pyproject.toml`
- Add an argument `--rich -r` in `fabric.py` to use rich in the normal output mode
- Add a function`rich_print()` to replace `print()` in `utils.py`

### Usage:
- Add `--rich` or `-r` to the commands.
- Not suitable for the stream output

### Demo 1: compare_and_contrast
demo1
### Demo 2: summarize
demo2
### Demo 3: create_coding_project
demo3


danielmiessler:

PR #718, opened 2024-07-10
Check for pattern files in working directory

CharlesHolbrow:

This enables fabric to find pattern files in the format `./<pattern-name>/system.md` in the working directory.  To generate an API request, fabric will first check the working directory, and then if no file is found, it will check the config directory. 

The main README recommends copying your custom patterns to the `[config]/patterns` dir. However, this is cumbersome when iterating on the pattern prompt, because you have to copy the patterns after each change.


danielmiessler:
Moving to go. Please resubmit if this isn't fixed after the migration.

PR #721, opened 2024-07-10
PROPOSED: Added caching for Youtube API (DONTMERGE)

mikewaters:
Proposed addition of on-disk caching for YouTube artifacts.

Not sure how useful it would be from a day-to-day Fabric usage perspective, but when iterating on a video-based Pattern, or comparing output from various models, it will reduce latency as well as YouTube API usage. 

It uses a newer library, [cachier](https://github.com/python-cachier/cachier), which when used in this way dumps pickle files to the user's `$HOME/.cachier`, and supports a few ways to define TTL. I think it works well in a cli/single-user use case, however I think there are better solutions if Fabric is used as a server. I am assuming, based on the location of `yt.py`, that the module is intended only for the local use case.

Profiler output:
```bash
           Time   Per Hit   % Time

From: 1132336.0    1e+06     63.1          video_response = get_video_details(youtube, video_id)
To:      1417.0   1417.0     14.8          video_response = get_video_details(youtube, video_id)

From: 651800.0 651800.0     36.3              transcript_list = get_transcript(video_id, languages=[options.lang])
To:      200.0    200.0      2.1              transcript_list = get_transcript(video_id, languages=[options.lang])
```
IMO it would need to be configurable, maybe even optional if one is primarily interested in Youtube comments.

Not sure if this aligns with your vision for the project. Also don't merge this, its messy.

### DIFF
--- a/installer/client/cli/yt.py
+++ b/installer/client/cli/yt.py
@@ -73,8 +73,6 @@ def get_comments(youtube, video_id):
 
 @profile
 def main_function(url, options, cachier__verbose=True):
-    print(url)
-    print(options)
     # Load environment variables from .env file
     load_dotenv(os.path.expanduser("~/.config/fabric/.env"))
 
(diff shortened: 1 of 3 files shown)


danielmiessler:
Moving to go. Please resubmit if this isn't fixed after the migration.

PR #802, opened 2024-08-12
Update patterns to contain user.md files

havok87:
Add missing user.md files to each pattern
## Related issues
This will resolve issues with using augmented_server.py and allow for consistent creation of pattern definitions


danielmiessler:
We're not using user.md.

PR #810, opened 2024-08-16
fix(ci): upload built binaries to GitHub Releases on tag creation

### LINKED ISSUE #808: [Feature request]: Provide binaries as part of releases
### What do you need?

It's awesome that you switched from Python to Go!  Now we don't need to have Python installed on our workstations to experience Fabric!

Since moving to Go, I assume the application is packaged as a single binary, and doesn't require the typical runtime dependencies (Python or Go), however your installation steps require Go to be installed on the machine.

It would be great if, as part of your release pipelines/processes, you could build and upload these built binaries for each platform so it's just a single curl + chmod to install Fabric. 

The main aim here is to remove as many of the dependencies the user needs to install to get up and running with Fabric, as it's something everyone should try out!

Apologies if I am missing some of the limitations that require this installation method.

If this is something you'd like to include, I am happy to work on it.

p5:
Updates the GitHub Actions workflow to be triggered on tag creation.
On tag creation, it runs the regular builds, but the only difference is a final step to upload the artifacts to GitHub Releases so they are discoverable.

Also manages the Go version in a single go.mod file, rather than across two different files, where the workflow version is easily forgotten. 

Finally updates the binary names to not include the distribution.  ubuntu-latest binaries work on Fedora, so therefore it makes more sense to me to call them `linux` and `darwin`.

## Related issues
Closes #808

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,14 +2,18 @@ name: Go Build and Release
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
+    tags:
+      - "v*"
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 
 jobs:
   build:
     name: Build binaries for Windows, macOS, and Linux
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
@@ -20,12 +24,12 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.22.1'
+          go-version-file: ./go.mod
 
       - name: Build binary on Linux and macOS
         if: matrix.os != 'windows-latest'
@@ -56,3 +60,12 @@ jobs:
           path: |
             fabric-${{ matrix.os }}-${{ matrix.arch }}-${{ github.ref_name }}
             fabric-macos-${{ matrix.arch }}-${{ github.ref_name }}.dmg
+
+      - name: Upload release artifact
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+        run: |
+          if [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            gh release upload ${{ github.ref_name }} fabric-macos-${{ matrix.arch }}-${{ github.ref_name }}.dmg
+          else
+            gh release upload ${{ github.ref_name }} fabric-${{ matrix.os }}-${{ matrix.arch }}-${{ github.ref_name }}*
+          fi
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -33,17 +33,21 @@ jobs:
 
       - name: Build binary on Linux and macOS
         if: matrix.os != 'windows-latest'
+        env:
+          OS: ${{ matrix.os == 'ubuntu-latest' && 'linux' || 'darwin' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -2,14 +2,18 @@ name: Go Build and Release
...
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 
 jobs:
   build:
     name: Build binaries for Windows, macOS, and Linux
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write
```
p5, line 16: Should provide the workflow with permissions to upload assets to GitHub releases.

.github/workflows/go.yml
```diff
@@ -20,26 +24,30 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
p5, line 27: Use a new Checkout action

.github/workflows/go.yml
```diff
@@ -20,26 +24,30 @@ jobs:
...
 
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.22.1'
+          go-version-file: ./go.mod
```
p5, line 32: Stop maintaining the Go version in two places.

.github/workflows/go.yml
```diff
@@ -20,26 +24,30 @@ jobs:
...
 
       - name: Build binary on Linux and macOS
         if: matrix.os != 'windows-latest'
+        env:
+          OS: ${{ matrix.os == 'ubuntu-latest' && 'linux' || 'darwin' }}
```
p5, line 37: Rebrand the `ubuntu-latest` binary to `linux`, and the `macos-latest` binary to `darwin`.

.github/workflows/go.yml
```diff
@@ -20,26 +24,30 @@ jobs:
...
         run: |
-          GOOS=${{ matrix.os == 'ubuntu-latest' && 'linux' || 'darwin' }} \
-          GOARCH=${{ matrix.arch }} \
-          go build -o fabric-${{ matrix.os }}-${{ matrix.arch }}-${{ github.ref_name }} .
+          go build -o fabric-${OS}-${{ matrix.arch }}-${{ github.ref_name }} .
 
       - name: Build binary on Windows
         if: matrix.os == 'windows-latest'
+        env:
+          OS: windows
```
p5, line 46: Rebrand the `windows-latest` binary to `windows`

.github/workflows/go.yml
```diff
@@ -56,3 +64,8 @@ jobs:
           path: |
             fabric-${{ matrix.os }}-${{ matrix.arch }}-${{ github.ref_name }}
             fabric-macos-${{ matrix.arch }}-${{ github.ref_name }}.dmg
+
+      - name: Upload release artifact
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+        run: |
+          gh release upload ${{ github.ref_name }} fabric-*-${{ matrix.arch }}-${{ github.ref_name }}*
```
p5, line 71: Upload the built files to GitHub Releases when invoked from a tag push.
Not entirely sure if this will work, but it will at least attempt to upload the artifacts.

.github/workflows/go.yml
ideologysec: strongly suggest removing the "put inside dmg" step on macos/darwin while you're at it

p5: As I'm not on the project, or involved in any way, I don't want to remove any functionality in this PR.  That's a decision for the maintainers.

I have, however, updated the pipeline to keep the MacOS binary around as well as the dmg in [3b35d88](https://github.com/danielmiessler/fabric/pull/810/commits/3b35d88611d8bfae01476c0e2a833db1e2300d0e).

PR #816, opened 2024-08-17
feat: added Base URL configuration to OpenAI

optman:
Base URL must be set for OpenAI-compatible service.

### DIFF
--- a/vendors/openai/openai.go
+++ b/vendors/openai/openai.go
@@ -30,6 +30,7 @@ func NewClientCompatible(vendorName string, configureCustom func() error) (ret *
 	}
 
 	ret.ApiKey = ret.AddSetupQuestion("API key", true)
+	ret.BaseURL = ret.AddSetupQuestion("Base URL", false)
 
 	return
 }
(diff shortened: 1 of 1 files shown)


vendors/openai/openai.go
```diff
@@ -30,18 +30,25 @@ func NewClientCompatible(vendorName string, configureCustom func() error) (ret *
 	}
 
 	ret.ApiKey = ret.AddSetupQuestion("API key", true)
+	ret.BaseURL = ret.AddSetupQuestion("Base URL", false)
```
eugeis, line 34: ret.BaseURL.Value = "https://api.openai.com/v1"

eugeis: Please set the default value at setting definition

eugeis (changes requested):
Please set the default value to "https://api.openai.com/v1"

vendors/openai/openai.go
optman: the default value will be utilized regardless if left unset.

eugeis: yes, that is clear. The user would know then, that a default values is available and he does not need to for one.

optman: I don't think the default value should be set here, instead it should be set in Openai package as it is now. If one day the Openai decide to upgrade from v1 to v2, we don't need to change this URL, only update Openai package is enough. 

I understand that you want to show a default value when the user run setup. But it is already marked as not required, so I think it's ok to leave empty.

Or we shouldn't ask users to set base url during setup. They can always set it manually as the YouTube API key.

Uddhav07:
as of now how can i change base url

danielmiessler:
We're going to skip this one because the default looks ok. If we're wrong please resubmit.

PR #872, opened 2024-08-23
fix: shadowing original error

ALX99:
This fixes shadowing the original error so that the original error is
propagated upwards

## Related issues
Shadowing original error

## Screenshots
N/A

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -56,7 +56,6 @@ func Init() (ret *Flags, err error) {
 	// takes input from stdin if it exists, otherwise takes input from args (the last argument)
 	if hasStdin {
 		if message, err = readStdin(); err != nil {
-			err = errors.New("error: could not read from stdin")
 			return
 		}
 	} else if len(args) > 0 {


cli/flags.go
```diff
@@ -56,7 +56,6 @@ func Init() (ret *Flags, err error) {
 	// takes input from stdin if it exists, otherwise takes input from args (the last argument)
 	if hasStdin {
 		if message, err = readStdin(); err != nil {
-			err = errors.New("error: could not read from stdin")
```
eugeis, line 59: err = fmt.Errorf("error: could not read from stdin: %w", err)

eugeis (commented):
Integrate the error label

cli/flags.go
ALX99: This is already covered in `readStdin` which wraps it: https://github.com/danielmiessler/fabric/pull/872/files#diff-4d5ebe4cc8e9de729d43f8f013c4b4a34bf473aed074c823c9178fb8f8938168L82

PR #895, opened 2024-08-30
Make the tweet prompt more of a prompt

ctwardy:
The "system.md" in patterns/tweet read more  like a Twitter HOWTO than a prompt. It even included how to create a Twitter account. 

I've kept some of the high-level flavor as it's an interesting approach, but cut the least relevant bits (like creating accounts), added markdown headings like other patterns, and some actual instructions.  I also added a section on Twitter Essays. 

## Notes
I've no claim to Twitter expertise so feel free to improve.
I suspect the LLMs all know about Twitter and this could be cut further.

### DIFF
--- a/patterns/tweet/system.md
+++ b/patterns/tweet/system.md
@@ -1,47 +1,39 @@
-Title: A Comprehensive Guide to Crafting Engaging Tweets with Emojis
+## Intent and Purpose
 
-Introduction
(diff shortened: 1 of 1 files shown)


danielmiessler:
I think we need to change the name of this thing and use a verb in front of it, and make it much more clear what it's doing.

PR #957, opened 2024-09-13
Add cmd -g to select the language would reply

quarechen:

Add a command -g to make the program return results according to the specified language.
Solve the multilingual adaptation problem. It can take prompts in various languages and return different results according to the same prompt.

## Screenshots
image


eugeis:
@quarechen Thank you for the PR.

Please change the default to be no language ("", means without additionally context text "please ..."), in order to avoid changes in the output.

quarechen:
> @quarechen Thank you for the PR.
> 
> Please change the default to be no language ("", means without additionally context text "please ..."), in order to avoid changes in the output.

ok

PR #984, opened 2024-09-20
adding flag for pinning seed in openai and compatible APIs

riccardo1980:
Using optional seed parameter in OpenAI API to get reproducible results 

## Changes
- adding -e flag to cli Flags
- propagating seed value up to go-openai.ChatCompletionRequest struct
- adding a log line (Debug level) to track system_fingerprint
- adding tests in openai and cli modules

## Related infos
- Even if we set the seed, we may get different results, we will have same results between two different calls only if Openai / Groq B/E status or model configuration is not changed. Changes are tracked using system_fingerprint parameter in return values.
- OpenAI relevant notes: https://platform.openai.com/docs/advanced-usage/reproducible-outputs
- Groq relevant notes: https://console.groq.com/docs/api-reference#chat

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ Application Options:
   -g, --language=                   Specify the Language Code for the chat, e.g. -g=en -g=zh
   -u, --scrape_url=                 Scrape website URL to markdown using Jina AI
   -q, --scrape_question=            Search question using Jina AI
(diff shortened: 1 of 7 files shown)


eugeis:
Thank you for PR. Please update also README.md with new flag.

PR #987, opened 2024-09-21
feat: remove cli list label and indentation

joshmedeski:
Trims the formatting on fabric -l, making it easier to combine with other CLI tools like fzf

## Before
SCR-20240921-gpou

## After
SCR-20240921-gpyb

## With gum (filtering CLI)
SCR-20240921-gqoi

This is a breaking change from the previous functionality. I also think we could combine this with a "raw" flag (ex: `fabric -l --raw`).

Let me know what you think, thanks!

### DIFF
--- a/db/storage.go
+++ b/db/storage.go
@@ -3,10 +3,11 @@ package db
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/samber/lo"
 	"os"
 	"path/filepath"
 	"strings"
+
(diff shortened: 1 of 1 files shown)


eugeis:
I like gum and would change the output in general, 
@danielmiessler, what’s your take? Should I go ahead and merge it?"

danielmiessler:
Yes I like this. We want to be as UNIXy as possible. Clean output for next input.

PR #996, opened 2024-09-26
add wipe flag for ctx and session

hallelujah-shih:
A wipe parameter is added to context and session for cleaning up outdated data

## Screen
Pasted image


eugeis:
Hi, thank you for the PR.
You don't need to implement the Wipe methods in Sessions and Contexts, the Delete method is already available inside of those structs by Storage struct, just call it.

hallelujah-shih:
@eugeis Thanks for the guidance, removed the Wipe method and called the Delete method of the Storage directly.

PR #1039, opened 2024-10-10
Feature/set default lang

hallelujah-shih:
Provide an option to set a default output language to make commands more concise.

## Screenshots

image

image


eugeis:
Thanks for the PR. Please update Readme.

hallelujah-shih:
> Thanks for the PR. Please update Readme.

The system will automatically set default values for you using --setup option, so there's no need to refer to the user manual.

PR #1088, opened 2024-10-27
feat: add DEFAULT_CONTEXT_LENGTH setting

### LINKED ISSUE #655: [Feature request]: Update patterns for better local LLM support (suggested fix)
### What do you need?

After some experimentation, q8 llama3 and mistral **run locally** are performing **much** better after adjusting the prompt order.  I propose restructuring the patterns so that the "# INPUT" text comes before output instructions.  INPUT before OUTPUT INSTRUCTIONS is producing much more coherent responses.

As an example, let's look at **extract_wisdom**:

# **Proposed** prompt order
INPUT
--put your YouTube transcript here--
IDENTITY and PURPOSE
STEPS
OUTPUT INSTRUCTIONS
OUTPUT

# **Current**  prompt order
IDENTITY and PURPOSE
STEPS
OUTPUT INSTRUCTIONS
INPUT

---------------------------------------------------------
Models tested:
llama3:8b-instruct-q8_0
mistral:7b-instruct-v0.2-q8_0

### LINKED ISSUE #662: [Question]: passing parameters to an ollama model
### What is your question?

Most local models have very small context windows.  There are a few which default to larger windows (128k and 256k), but most default to very small context size.

Several models offer a configurable parameter to set the limit, but I don't see any method to pass that parameter through the CLI.

Example: llama3-gradient can extend the size of num_ctx from the CLI with this command: 

/set parameter num_ctx 256000

https://ollama.com/library/llama3-gradient

Is there, perhaps, an undocumented CLI switch to allow setting num_ctx the same way as --temp and --top_p ??

### LINKED ISSUE #743: [Bug]: Local LLM Output Short and Unformatted Remarks Instead of Following Prompt instructions and output guidlines with GPT-4, Claude, etc #bug #question
# Local LLM Output Short, Unstructured and Unformatted Compared to GPT-4

#### Description:
Users have reported that when using the models locally, the output is significantly shorter and does not follow the prompt instruction or the formatting provided by GPT-4. This issue is consistent across different setups, and different local models including various hardware configurations and operating systems.

#### Reproduction Steps:
1. Use `yt --transcript` command with a YouTube video link.
2. Run `fabric -m llama3:latest -sp extract_wisdom`.
3. Compare the output with the same process using GPT-4.

#### Observed Behavior:
- Local LLM produces 5-15 sentences without formatting or following the prompt instructions.
- GPT-4 creates comprehensive outputs with insights, quotes, and recommendations.

#### Expected Behavior:
Local LLM should produce similar comprehensive outputs as GPT-4, or similar subpar responses.

#### Additional Context:
- Various hardware configurations: NVIDIA 4090, GTX 1070, RTX 3080, 3060.
- Operating systems: Arch Linux, Windows 10 WSL.
- Parameters tried: temperature, top_p, repeat_penalty, num_ctx.

#### Discussion Links:
- #504  [Discussion 504](https://github.com/danielmiessler/fabric/discussions/504)
- #428 [Discussion 428](https://github.com/danielmiessler/fabric/discussions/428)
- #740  [Question 740](https://github.com/danielmiessler/fabric/issues/740)

jaredmontoya:
This PR adds support for setting default Model Context Length.

Motivation:
- I don't want to create a separate extended context model in Ollama just to set the num_ctx parameter, also you can't change it on the fly if you do so.
- This will fix all issues that are caused by Ollama's default 2048 token context window. They are common because Fabric's best patterns like `extract_wisdom`, `summarize` and others almost always require high context window lengths to work on large inputs.

This PR adds:
- `--contextLength` flag to set context length using cli
- `DEFAULT_CONTEXT_LENGTH` option to `.config/fabric/.env` (used if `--contextLength` is not used)
- *if neither `--contextLength` nor `DEFAULT_CONTEXT_LENGTH` are set context parameter is not sent at all*
- `--dry-run` support is included

How it works:
- Now when choosing the default model during setup you can optionally specify the context window length to use with it because the model size and the context window you can afford corelate a lot.
- The parameter is not sent at all if left blank which lets Ollama later implement dynamic context window adjustment based on your VRAM.
- I chose to add context length to global defaults instead of ollama defaults to simplify implementation of the `--dry-run` vendor and so that in the future it can be reused for other vendors if they ever provide an API option to change the context window length. As well as because of the corelation between model size and how much more memory you can give to it's context window.

Notes:
- I realize that context length is a term that fights with fabric's context concept and that fabric already used all possible short one-letter flags that relate to the word context
- I do not provide a short one-letter cli flag for setting contextLength because there are none left that mean context and because it's more of a debugging feature anyway. Users will just use the `.env` config to set this value.

## Related issues
*actually* closes #662
*actually* closes #743
*possibly* closes #655

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -2,16 +2,17 @@ package cli
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
 	"github.com/danielmiessler/fabric/common"
 	"github.com/danielmiessler/fabric/core"
 	"github.com/danielmiessler/fabric/plugins/ai"
 	"github.com/danielmiessler/fabric/plugins/db/fsdb"
 	"github.com/danielmiessler/fabric/plugins/tools/converter"
 	"github.com/danielmiessler/fabric/restapi"
-	"os"
-	"path/filepath"
-	"strconv"
-	"strings"
 )
 
 // Cli Controls the cli. It takes in the flags and runs the appropriate functions
@@ -211,7 +212,7 @@ func Cli(version string) (err error) {
 	}
 
 	var chatter *core.Chatter
-	if chatter, err = registry.GetChatter(currentFlags.Model, currentFlags.Stream, currentFlags.DryRun); err != nil {
+	if chatter, err = registry.GetChatter(currentFlags.Model, currentFlags.ModelContextLength, currentFlags.Stream, currentFlags.DryRun); err != nil {
 		return
 	}
 
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -4,14 +4,15 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
-	goopenai "github.com/sashabaranov/go-openai"
 	"io"
 	"os"
 	"strings"
 
-	"github.com/danielmiessler/fabric/common"
 	"github.com/jessevdk/go-flags"
+	goopenai "github.com/sashabaranov/go-openai"
 	"golang.org/x/text/language"
+
+	"github.com/danielmiessler/fabric/common"
 )
 
 // Flags create flags struct. the users flags go into this, this will be passed to the chat struct in cli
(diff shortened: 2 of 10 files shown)


jaredmontoya:
@eugeis I saw that you merged my other PR 12 minutes ago, I fixed merge conflicts in this one so it can also be merged.

eugeis:
@jaredmontoya I like you PR. 
Would you like to change the name to contextWindow or something like this. Because user will be confused, because "your" context is something else, then the prompt context.

eugeis:
Please update also the gomod2nix.toml file, because I updated dependencies in last commits.

jaredmontoya:
> Did this land in a release yet? I'm running `v1.4.122` and calling `fabric --contextLength` returns
> 
> ```
> unknown flag 'contextlength'
> ```

I was required to rename it to `--modelContextLength`

morrolinux:
@jaredmontoya oh, you're right, I'm sorry. Thanks for the hint anyways :)

PR #1089, opened 2024-10-27
Introduce Nix to the project

jaredmontoya:
*Don't run `extract_wisdom` or `summarize` on this*

I am the maintainer of `fabric-ai` package in the central NixOS repository called [nixpkgs](https://github.com/NixOS/nixpkgs)

Earlier I created a PR #1088, but to make development a lot more comfortable for myself and other Nix users who decide to contribute to fabric I also wrote this Nix flake which will bring benefits of the Nix ecosystem to fabric.

It improves developer experience and lets people unclutter their systems. Flake maintenance is not resource intensive at all if you know what you are doing, so any Nix user can create a PR if `flake.lock` needs updating.

What does this bring to the project:
- Reproducibility, if you use `nix build` to build fabric the build will be binary level reproducible on any system because everything is pinned in the `flake.lock` including the compiler.
- Development shell, this flake provides a development shell configuration that can be activated automatically using [direnv](https://github.com/direnv/direnv) when entering the project root. It includes everything that you will need to develop the project including the `go` compiler, go tools, gomod2nix, 
and the helper script `update` that basically executes `go mod -u && go mod tidy && gomod2nix generate`. Everything in this flake is pinned using `flake.lock` which is checked into git so you also get a history of ALL your dependencies including the compiler, not just go modules.
-  This essentially means that even in a lot of years when old fabric code is incompatible with the new go compiler version you can just revert to the old git commit and automatically get the old compiler version and you will be able to build old versions of fabric with no problems.
- **The package provided by this `flake.nix` means that NixOS users can directly install `fabric` from your git repository and any of it's branches or tags bypassing the `fabric-ai` package the central [nixpkgs](https://github.com/NixOS/nixpkgs) repository that is updated a lot less frequently.**
- `nix fmt` formats all go files codebase with `gofmt` and all nix files with various nix formatters. It is possible to add `yamlfmt.enable = true;` to `treefmt.nix` and then it will also format yaml files, basically it's a tool to format any text files in the project. `nix flake check` can be used to check  if all files are formatted properly. In the future I probably will open a PR that adds `goimports-reviser` support to `treefmt-nix` so it will also be able to sort imports.

If you don't know what Nix is and don't use it:
- You can just ignore all files in this PR and use your own compiler like you did before even with those files present. Nothing changes for you personally.
- In the future, if you need to build an old version of fabric with an old version of the compiler instead of having only one option you will have two options:
- - Find out which version of the go compiler is needed for that commit, find it's package somewhere, install it, build the commit you want.
- - Install Nix, checkout the needed commit, run either `nix build` for a production package or `nix develop` to enter a devshell with the needed compiler and run `go build`

What to keep in mind:
- Unfortunately if you run `go get -u && go mod tidy` or modify `go.mod` in any other way and then don't run `gomod2nix generate` after that it will desync `go.mod` and `gomod2nix.toml` which can lead to unexpected results when using `nix build`or installing the nix package, **normal builds will not be affected**.
- **This problem can be mitigated by creating a github workflow that runs `gomod2nix gnerate` on all commits so that if `go.mod` is changed in any of them `gomod2nix.toml` is automatically synced.**
- Another solution might be to just install [gomod2nix](https://github.com/tweag/gomod2nix) and remember to run it whenever you change `go.mod` but you will have to remind this to everyone who creates a PR that requires changing `go.mod`


jaredmontoya:
For Nix users: If you want to get the context length fix from #1088 before it is even merged you can do this by adding an overlay like this:
```nix
inputs = {
  fabric = {
        url = "github:jaredmontoya/fabric";
        inputs = {
          nixpkgs.follows = "nixpkgs";
        };
      };
};
```
```nix
    nixpkgs.overlays = [
      (self: super: {
        fabric-ai = inputs.fabric.packages.${system}.default;
      })
    ];
```

PR #1096, opened 2024-10-30
Implement automated ci nix package version update

jaredmontoya:
Implemented automatic CI updates of the nix package version


jaredmontoya:
@eugeis sorry for another PR, but I remembered that a github workflow updates version.go and that it will always be out of sync with the nix package version if I don't implement this in the CI.

PR #1126, opened 2024-11-09
flake: add gomod2nix auto-update

jaredmontoya:
Previously I added a nix flake for this project, but it requires constant `gomod2nix.toml` maintenance and as a result the reproducible fabric package provided by the flake broke when `go.mod` was updated without syncing `gomod2nix.toml`

This pull request adds automatic gomod2nix updates to the existing github workflow.


eugeis:
Hi @jaredmontoya, 
we need restructure file system, would you like to do it for Nix?

See #1127

Issue #1144, opened 2024-11-17
[Feature request]: Enhanced Variable Substitution and Input Processing


mattjoyce:
### What do you need?

# Feature Request: Enhanced Variable Substitution and Input Processing in Fabric

## Overview
Propose improvements to Fabric's variable handling and input processing to enhance automation capabilities while maintaining backward compatibility through configuration.

### 1. Standardized Variable Notation
Move from direct substitution to double brace notation with simplified CLI usage.
{{role}}
{{max_word_count}}

#### Why Double Braces?
1. Industry Standard Template Syntax:
   - Used in many popular templating systems
   - Immediately recognizable to developers from various backgrounds
   - Clear visual separation between variables and content
   - Less prone to accidental matches than simpler markers like # or $

2. Advantages over alternatives:
   - `#variable`: Can conflict with Markdown headers, hashtags
   - `$variable`: Could be confused with shell variables, environment variables
   - `%variable%`: Less common in modern systems
   - Single braces: Could conflict with JSON/YAML content

3. Better Pattern Readability:
   ```
   # Current syntax (less clear):
   You are a #role evaluator with #points maximum points.

   # Proposed syntax (clearly delineated):
   You are a {{role}} evaluator with {{points}} maximum points.
   ```

4. Future-proofing:
   - Double brace syntax enables powerful plugin architecture integration
   - Nested variable resolution: inner variables resolved before plugin processing
   - Example plugin capabilities:

     a. System Information:
     ```
     {{plugin:sys:hostname}}
     {{plugin:sys:user}}
     ```

     b. Date and Time Operations:
     ```
     {{plugin:date:now:RFC3339}}
     {{plugin:date:today:YYYY-MM-DD}}
     {{plugin:date:tomorrow:DD/MM/YYYY}}
     {{plugin:time:utc}}
     ```

     c. Text Processing (with variable nesting):
     ```
     {{plugin:text:trim:{{name}}}}
     {{plugin:text:uppercase:{{name}}}}
     ```

     d. Data Operations:
     ```
     {{plugin:yaml:format:{{config}}}}
     ```

     e. Database Queries (with variable interpolation):
     ```
     {{plugin:sqlite:users.db:SELECT * FROM users WHERE name={{username}}}}
     {{plugin:postgres:SELECT count(*) FROM logs WHERE date > {{date}}}}
     ```

   These examples demonstrate how the double brace syntax enables:
   - Simple plugin calls
   - Nested variable resolution
[...]

PR #1147, opened 2024-11-18
refactor: unify pattern loading and variable handling

mattjoyce:
- Stronger separation of concerns between chatter.go and patterns.go
- Consolidate pattern loading logic into GetPattern method
- Support both file and database patterns through single interface
- Maintain API compatibility with Storage interface
- Handle variable substitution in one place
- Keep backward compatibility for REST API through Get method

The changes enable cleaner pattern handling while maintaining existing interfaces and adding file-based pattern support.


mattjoyce:
#1144 will probably significantly further refactor this in the future.

PR #1157, opened 2024-11-22
fix: process template variables in raw input

mattjoyce:
Process template variables ({{var}}) consistently in both pattern files and raw input messages. Previously variables were only processed when using pattern files.

- Add template variable processing for raw input in BuildSession
- Initialize messageContent explicitly
- Remove errantly committed build artifact (fabric binary in previous commit)

This will now support inline macros, with missing var checks.
 echo 'Hello {{name}}' | ./fabric -v=name:World


ksylvan:
@mattjoyce Please provide a way to turn off this new behavior (preferably make it so you have to turn it on with a new flag, like `--use-templates`)

It's a great feature, but it breaks existing workflows where you're feeding files (or snippets of files, like diffs) that us Go templates or Jinja2 format; e.g Ansible Playbooks.

PR #1174, opened 2024-12-01
Curly brace templates

mattjoyce:
Fixes segfault iss if pattern is used with no user input.

./fabric -p foo


mattjoyce:
fixed up some merge conflicts.  rebased my feature branch with upstream main.

PR #1189, opened 2024-12-10
Add --input-has-vars flag to control variable substitution in input

mattjoyce:
Make template variable substitution in input opt-in via --input-has-vars flag

## Problem
When users paste Ansible, Jekyll, or other content containing {{variable}} syntax into fabric, 
our template processor attempts to resolve these variables, leading to unwanted variable 
substitution and errors about missing variables.

## Solution
- Added new --input-has-vars flag
- Template variable substitution in user input only occurs when flag is set
- All other template processing (patterns, etc.) remains unchanged
- Preserves literal curly braces by default

## Testing
```bash
# Without flag - preserves template syntax
echo "Hello my name is {{name}}" | ./fabric -v=name:Matt
> Hello! It looks like there might have been a small error...

# With flag - performs substitution
echo "Hello my name is {{name}}" | ./fabric -v=name:Matt --input-has-vars
> Hello Matt! How can I assist you today?

Relates to : 
https://github.com/danielmiessler/fabric/issues/1184
https://github.com/danielmiessler/fabric/issues/1171

Issue #1184, opened 2024-12-09
[Bug]: jinja2 syntax in input text causes errors


ksylvan:
### What happened?

When passing a text that contains Jinja2 syntax to the AI, it looks like it tried to be interpolated, resulting in an error.

This makes it impossible to use, for example, diffs when working with Go templates or Ansible playbooks as input to `summarize_git_diffs`

### Version check

- [X] Yes I was.

### Relevant log output

```shell
fabric --version
v1.4.119


Here is a simple use case:

```text
$ ai '{{ foo.bar }}'
missing required variable:  foo.bar 
```

ksylvan:
This is likely due to https://github.com/danielmiessler/fabric/pull/1157

@mattjoyce We need a new command-line flag to enable the new variable substitution behavior so that people who work with diffs or text files that contain go template or Jinja2 syntax can still use fabric patterns.

mattjoyce:
@ksylvan yep, I totally missed this use case.

@eugeis , I think we should perhaps implement --inputvars, which will do substitution on input content, default to not.

ksylvan:
@mattjoyce Thanks, that's perfect.

ksylvan:
There's still a slight bug, @mattjoyce 

```text
$ git diff | summarize_git_diff
could not get pattern summarize_git_diff: missing required variable:  domain 

$ fabric --version             
v1.4.120

$ fabric --help | grep input-has
      --input-has-vars      Apply variables to user input
```

Fixed with this patch, @mattjoyce 

```diff
diff --git a/plugins/db/fsdb/patterns.go b/plugins/db/fsdb/patterns.go
index 968b08d..6fedb72 100644
--- a/plugins/db/fsdb/patterns.go
+++ b/plugins/db/fsdb/patterns.go
@@ -68,7 +68,7 @@ func (o *PatternsEntity) applyVariables(
        }
 
        var result string
-       if result, err = template.ApplyTemplate(pattern.Pattern, variables, input); err != nil {
+       if result, err = template.ApplyTemplate(pattern.Pattern, variables, ""); err != nil {
                return
        }
        pattern.Pattern = result
```

mattjoyce:
I'm not sure that's right.  With out that the user input will not be appended or inserted in to the final context.
I think I'll have to hardcode something for that specific var {{input}}, and now I think of it, input is a bit generic, and plausible a diff or other template will use it.  will probably change it fabric_user_input.

ksylvan:
[...]

Issue #1171, opened 2024-11-29
[Bug]: Running `improve_prompt` on a prompt with template variables fails


osheari1:
### What happened?

With a prompt, such as `improve_security_policy`

```markdown
# IDENTITY and PURPOSE
You are an expert in cybersecurity standards and policies. You specialize in creating clear, well written policies for companies and corporations (for example for certifications such as SOC2 and HiTrust).

You have been tasked with improving an existing security policy that needs to be improved, and you must output an improved report finding in markdown format.

# GOAL
Given a policy type and set of requirements, provide a well written, concise improved security policy in markdown format.

# STEPS
...
# OUTPUT INSTRUCTIONS
...
# INPUT
INPUT:
## Requirements: 
{{reqs}}
## Policy 
```

piping into `improve_prompt` returns

```bash
$ cat system.md | fabric --pattern improve_prompt 
missing required variable: reqs
```

Expected behavior:
Improve prompt would work on patters with variables / plugins / other future stuff that may be added to the templating features. 

I'm assuming this is because variables are validated after stdin is substituted into the template. 

### Version check

- [X] Yes I was.

mattjoyce:
hello @osheari1 
yes, you're correct.

the user input and pattern are combined an then the template system tries to resolve is double brace pairs. {{ }}
A work around might be just to provide a variable.  -v=req:""

osheari1:
Hello!

I'm wondering if in the long run, a more robust solution would be needed... :thinking: 

Thinking of two cases:
- Prompts or INPUT that contains non-fabric variable substitution `{{...}}` syntax (eg. Jekyll and other template libs),
- INPUT that contains fabric variable substitutions that we don't want substituted

Pattern
```markdown
# Example Output
// Some example explaining Jekyll syntax
{{ jekyll-var-no-sub-wanted }}

# Output 
{{ some-fabric-variable }} 
```

Input
```markdown
Stuff {{ fabric-var-no-sub-wanted }}
```
Here, `{{ fabric-var-no-sub-wanted}}` and `{{ jekyll-var-no-sub-wanted }}` would both throw errors. 

**Problem**
- All `{{...}}` syntax in prompt and input are substitution targets. Aka fabric doesn't know which variable should require substitution.
  - To users of a very complex prompt with multiple non-fabric and fabric var substitutions, it may get confusing as to which substitutions you actually want to substitute. 
- Variable substitution happens on both prompt and input at the same time.

**Questions**
[...]

PR #1197, opened 2024-12-12
bug: Using patterns, fabric needlessly scans input for template vars

ksylvan:

Fix a minor bug related to https://github.com/danielmiessler/fabric/pull/1189

## Related issues

#1171 
#1184 

## Result

Using the simplest pattern (`ai`):

```text
$ echo 'My name is {{ name }}' | fabric -p ai   
could not get pattern ai: missing required variable:  name
```

And after the fix:

```text
$ echo 'My name is {{ name }}' | fabric -p ai  
- Your name is a core part of your identity.
- It can influence personal and social interactions significantly.
- Names often carry cultural, familial, or historical significance.
- They can shape perceptions and first impressions in various contexts.
- Understanding your name's meaning can offer personal insights.
```

And variable substitution:

```text
$ echo 'My name is {{ name }}' -vname=Kayvan --input-has-vars | fabric -p ai   
- Hello, Kayvan! How can I assist you today?  
- Do you have any specific questions or topics in mind?  
- Feel free to share any details or context for assistance.  
- I'm here to help with insights or advice you need.  
- Let's explore solutions or ideas tailored to your interests.
```

### DIFF
--- a/core/chatter.go
+++ b/core/chatter.go
@@ -130,8 +130,12 @@ func (o *Chatter) BuildSession(request *common.ChatRequest, raw bool) (session *
 
 	var patternContent string
 	if request.PatternName != "" {
-		pattern, err := o.db.Patterns.GetApplyVariables(request.PatternName, request.PatternVariables, request.Message.Content)
-		// pattrn will now contain user input, and all variables will be resolved, or errored
+		contentToPass := ""
+		if request.InputHasVars {
+			contentToPass = request.Message.Content
+		}
+		pattern, err := o.db.Patterns.GetApplyVariables(request.PatternName, request.PatternVariables, contentToPass)
+		// pattern will now contain user input, and all variables will be resolved, or errored
 
 		if err != nil {
 			return nil, fmt.Errorf("could not get pattern %s: %v", request.PatternName, err)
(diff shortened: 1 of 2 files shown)


mattjoyce:
I think this will break patterns containing {{input}}.   This specially added so user input could be positioned within the prompt, not appended.

ksylvan:
@mattjoyce Good point. What's the right way to fix this? Basically, we want the ability to specify `{{input}}` in pattern files while leaving the input completely alone if the `--input-has-vars` flag is not passed in by the user.

mattjoyce:
Yep, I think we have to rename {{input}} to something unique (I should have
anticipated that), then in template.go check for it, and not recurse into
that pair of curly braces.

I will have a look tonight.

ksylvan:
That would work, yes. Call it something like `{{__fabric_prompt_input__}}` or something equally  long and unique.

Then, if `--input-has-vars` is not set, we only have to look for and replace that special, built-in variable, and if it is set, we do the full templating like we are doing now.

mattjoyce:
In theory yes but the template system runs until there are no unresolved
curly braces pairs.  And they can be embedded like this {{outer{{inner}}}}
so not resolving input will need some other changes.

mattjoyce:
Ok, I think fixed in this PR - much simpler then I anticipated.
https://github.com/danielmiessler/fabric/pull/1200

ksylvan:
Better fix in #1200

PR #1208, opened 2024-12-16
Fix: Issue with the custom message and added example config file.

mattjoyce:
This fixes an issue introduced in the last feature, where the custom message was incorrectly handled.
In this usage 
echo "foo" | ./fabric --pattern ai "custom message" --dry-run

it would erroneously append --dry-run to the user input.

Added example.yaml and readme

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,0 +1,68 @@
+# YAML Configuration Support
+
+## Overview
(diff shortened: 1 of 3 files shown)


mattjoyce:
@eugeis does this need any discussion?

PR #1219, opened 2024-12-22
Update translate pattern to use curly braces

iqbalabd:
Updates the pattern for translation to use curly braces so we can insert the target language as a parameter on the command line, i.e

`$ cat my-english-text.md | fabric -v=lang_code:ja-jp -p translate`


mattjoyce:
The language can be specified as a cli options already.
 -g, --language=           Specify the Language Code for the chat, e.g. -g=en -g=zh

It becomes appended to the system role message.

iqbalabd:
> The language can be specified as a cli options already. -g, --language= Specify the Language Code for the chat, e.g. -g=en -g=zh
> 
> It becomes appended to the system role message.
> 
> I'm not saying using it in a pattern is wrong - just sharing.

Thanks. My understanding of the `--language=` parameter is that it changes the whole chat language, although it doesn't specify the language to translate to for this particular pattern. Is that correct?

eugeis:
Somehow I can't merge this PR. 
There were wrong version files, I deleted them and now no merge possible. 
Please create new PR (without version files).

iqbalabd:
For some reason those version files got included in the PR although I did not commit them explicitly. I couldn't figure out why, but I did notice the differences in those files when I compared with the upstream version. Syncing with upstream doesn't seem to work.
So I ended up removing my fork, re-forked it again and this time it worked. [Here](https://github.com/danielmiessler/fabric/pull/1230) is the new PR.

PR #1222, opened 2024-12-22
Fix cross-filesystem file move in to_pdf plugin (issue 1221)

wmahfoudh:
This PR resolves a bug in the `to_pdf` plugin where it fails to move files across different filesystems. The issue arises from the use of the `os.Rename` function, which does not support cross-device file moves. The fix replaces `os.Rename` with a combination of `io.Copy` and `os.Remove`, ensuring compatibility with environments where /tmp and the current working directory are on different filesystems.

## Related issues
closes #[1221]

## Screenshots
None


Tarekinh0 (approved):
👌

PR #1246, opened 2025-01-05
New pattern enhanced write latex

wmahfoudh:
This pull request proposes an enhanced version of the write_latex as a new pattern (write_latex_enh). The pattern specifies additional LaTeX rules that improve the output code. For instance it improves support for accented characters like in French language, as well as double and single quotes. Code generated is less prone to break when converted to pdf using to_pdf.
I did not want to replace the original pattern as this gives more options to the user.

## Related issues
No related issues.


eugeis:
Thank you for the PR.
Please remove version* files.

wmahfoudh:
HI, 
I deleted the wrong file :)
Will create another request.

PR #1349, opened 2025-03-08
Fix: v1.4.153 does not compile because of extra version declaration

ksylvan:
## Related issues
N/A

### Before fix

Fabric fails to compile:

```plain
$ go install .
# github.com/danielmiessler/fabric
./version.go:3:5: version redeclared in this block
        ./main.go:12:5: other declaration of version
```

The `version` string is set up in `version.go` and does not need to be declared again

```go
package main

var version = "v1.4.153"
```

### After fix

```plain
kayvan@dharma src/fabric [03-08-extra-version-declaration-removed] $ go install .
kayvan@dharma src/fabric [03-08-extra-version-declaration-removed] $ fabric --version
v1.4.153
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,8 +9,6 @@ import (
 	"github.com/danielmiessler/fabric/cli"
 )
 
-var version string
-
 func main() {
 	err := cli.Cli(version)
 	if err != nil && !flags.WroteHelp(err) {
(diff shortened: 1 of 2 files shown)


ksylvan:
@eugeis @danielmiessler This PR also fixes the tests that are failing. See for example, https://github.com/danielmiessler/fabric/pull/1350/checks

jmd1010:
That probably explains why my pull request test is failing:)  due to double declaration of go.

ksylvan:
> That probably explains why my pull request test is failing:) due to double declaration of go.

Yes @jmd1010 - That and the Azure client fix.

PR #1363, opened 2025-03-15
clipboard operations now work on Mac and PC

garkpit:
Makes copy/paste in the streamlit app work across all platforms (Mac, PC, Linux)

## Related issues
I didn't check

## Screenshots
None

ps. 
1 - This is the first time I've ever contributed to anything. Please be kind! Let me know if I've done anything wrong :)
2 - Done using Claude 3.7 Thinking in Cursor
3 - Tested on Mac. Not tested on PC or Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,9 +582,9 @@ The `web/src/lib/content` directory includes starter `.obsidian/` and `templates
 
 ### Installing
 
-The GUI can be installed by navigating to the `web` directory and using `npm install`, `pnpm install`, or your favorite package manager. Then simply run the development server to start the app.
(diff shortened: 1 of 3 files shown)


eugeis:
Hi @garkpit, thank you for the PR! We're aiming to keep the solution as close to pure Go as possible. If I understand correctly, your PR introduces a Python-based approach. Would you be open to providing a Go-based solution or an OS-native alternative that doesn’t rely on Python dependencies?

garkpit:
Hi @eugeis 
The fix is just for the streamlit app (which is python). If the requirements.txt is confusing (I can see why) then we can just remove it and update the readme accordingly.

PR #1399, opened 2025-03-31
feat: add simple optional api key management for protect routes in --serve mode

HaroldFinchIFT:
Hi all. This PR introduces basic API Key protection for the `--serve` mode in Fabric. Currently, when running Fabric with `--serve`, the API is completely open, which poses security risks — especially considering that `GET /config` exposes API keys of various AI models.

### Motivation
- I wanted to deploy Fabric as a Docker container at home exposed in my network in `--serve` mode.
- Without network-level security configurations, the API would be freely accessible within the network.
- This change is a first step towards improving security, providing a simple way to restrict access.

### Implementation
- Added an `--apikey` flag to protect server routes.
- If set, requests must include `X-API-Key: <API_KEY>` header.
- If not set, behavior remains unchanged for backward compatibility.
- Implemented as a middleware (auth.go) that checks the request header before allowing access to routes.

### Testing
- I tested this locally, and the behavior seems to work as expected.

This is just an initial proposal, open to discussion and improvements. I’m always mindful of security, so I thought this could be a useful contribution. I’m not a Go expert, so feedback and constructive criticism are very welcome!

Looking forward to your thoughts!
Thanks!


ksylvan:
Hi @HaroldFinchIFT - you can achieve the same protection without altering Fabric’s code by placing a reverse proxy (or API gateway) in front of Fabric’s “–serve” endpoint.

For example, you can configure NGINX, Traefik, Caddy, or HAProxy to inspect incoming HTTP requests and require a valid API key (such as via an “X-API-Key” header) before forwarding the request to Fabric. This approach is generic, keeps your Fabric installation unchanged, and lets you add extra security layers like TLS termination, rate limiting, or logging.

For example, something like this `compose.yml` file with that runs a Caddy proxy:

```yaml
version: "3.8"
services:
  fabric:
    # CHANGE the below to your compiled fabric docker image and add whatever volumes you need
    # for the .config/fabric directory...
    image: your_user/fabric:latest
    command: ["fabric", "--serve"]
    expose:
      - "8080"  # expose port 8080 to other containers
    restart: unless-stopped

  caddy:
    image: caddy:2
    ports:
      - "80:80"
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile:ro
    environment:
      API_KEY: "supersecret"  # set your desired API key here
    depends_on:
      - fabric
```

And the corresponding Caddyfile:

```cady
:80 {
    @authorized {
        header X-API-Key {env.API_KEY}
    }
    handle @authorized {
        reverse_proxy fabric:8000
    }
    handle {
        respond "Forbidden" 403
    }
}
```

Then you should be able to do:

```bash
docker compose up -d
```

This solution adds a generic security layer without modifying Fabric’s code.

HaroldFinchIFT:
Thank you for your feedback! I understand the reasoning behind using a reverse proxy for API key authentication without modifying Fabric, but I’d like to clarify why I believe an internal API key check is still valuable:

**Direct accessibility concerns** – A reverse proxy doesn’t fully solve the issue in all environments. In my setup, Fabric runs inside a Docker network using an IPvLAN L3 configuration. In this scenario, other containers and devices on the network can still access Fabric directly, bypassing the proxy. Securing this requires additional firewalling and network segmentation, adding complexity.

**Security by default** – While a proxy can enhance security, not all users will or can configure one. This change introduces a simple, built-in security mechanism that doesn’t interfere with proxy-based setups but provides a lightweight option for those who prefer an easier approach. The implementation remains entirely optional—if `--apikey` is not set, Fabric behaves exactly as before.

**Future scalability** – This is a minimal first step but could lay the groundwork for more structured access control in the future. Given that Fabric now has a GUI, multi-user support might become relevant—for example, in small teams or families wanting shared private patterns while keeping responses segmented. Having a first internal authentication mechanism could facilitate this.

Would you be open to leaving this PR open for discussion to see if others find this feature useful? If there’s interest, I’d be happy to adjust the implementation based on feedback.

Looking forward to your thoughts, and thanks again for your time and feedback!

ksylvan:
> Would you be open to leaving this PR open for discussion to see if others find this feature useful? If there’s interest, I’d be happy to adjust the implementation based on feedback.

I don't really have the power to close a PR, not unless @eugeis or @danielmiessler grant me maintainer privs here. 😄 

Let me look carefully at the PR and give my purely code-focused review.

cli/flags.go
```diff
@@ -64,6 +64,7 @@ type Flags struct {
 	Serve                           bool              `long:"serve" description:"Serve the Fabric Rest API"`
 	ServeOllama                     bool              `long:"serveOllama" description:"Serve the Fabric Rest API with ollama endpoints"`
 	ServeAddress                    string            `long:"address" description:"The address to bind the REST API" default:":8080"`
+	ServeApiKey                     string            `long:"apikey" description:"API key used to secure server routes" default:""`
```
ksylvan: I suggest changing this to:

```go
ServeAPIKey string `long:"api-key" description:"API key used to secure server routes" default:""`
```

- Use `ServeAPIKey` rather than `ServeApiKey` to match the standard Go naming convention for acronyms (e.g., "API", "HTTP").
- Using a hyphen (`api-key`) in CLI arguments is a common convention that improves readability and consistency across CLI tools.

restapi/auth.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+func ApiKeyMiddleware(apiKey string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		headerApiKey := c.GetHeader("X-API-Key")
```
ksylvan: rename this to use `API` instead of `Api` (APIKeyMiddleware, etc.)

restapi/auth.go
```diff
@@ -0,0 +1,21 @@
...
+
```
ksylvan, line 19: I suggest adding this snippet here:

```go
if headerAPIKey == "" {
       c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "Missing API Key"})
            return
}
```

restapi/serve.go
```diff
@@ -5,13 +5,17 @@ import (
...
+func Serve(registry *core.PluginRegistry, address string, apiKey string) (err error) {
 	r := gin.New()
 
 	// Middleware
 	r.Use(gin.Logger())
 	r.Use(gin.Recovery())
 
+	if apiKey != "" {
+		r.Use(ApiKeyMiddleware(apiKey))
+	}
```
ksylvan, line 21: Maybe we should add a warning here if the API key is not set?

```go
if apiKey != "" {
        r.Use(APIKeyMiddleware(apiKey))
} else {
        log.Warn("Starting REST API server without API key authentication. This may pose security risks.")
}
```

ksylvan (changes requested):
- Consider making the header name configurable or at least define it as a constant to avoid hardcoding it in multiple places.

```go
const APIKeyHeader = "X-API-Key"
```

then use `APIKeyHeader` in the places where you hard-code `"X-API-Key"`

cli/flags.go
HaroldFinchIFT: Ok, done! Thanks for the suggestions. I didn't know about Go acronyms naming convention!

restapi/auth.go
HaroldFinchIFT: Same above.

restapi/auth.go
HaroldFinchIFT: Good point! I updated the code to explicitly handle the missing API key case separately. Thanks

restapi/serve.go
HaroldFinchIFT: Ok, it makes sense. I've done also this. Thanks

ksylvan (approved):
The changes look good to me. If I had the power, I'd merge this. 👍

PR #1405, opened 2025-04-11
Fix chat response sequence in Web UI

jmd1010:
Address unwanted behavior where sometimes the new llm response does not automatically post  below the previous exchange in the chat history svelte interface.


jmd1010:
sorry pulled  from wrong branch. Will resubmit

jmd1010:
as per above

PR #1420, opened 2025-04-17
Fix error in deleting patterns due to non empty directory

sherif-fanous:

A Fabric pattern is made up of a directory in `~/.config/fabric/patterns` where the name of the directory represents the pattern name. Within that directory are one or more `.md` files that describe the pattern.

With the current implementation that uses `os.Remove`, invoking the `DELETE /patterns/:name` API results in the following error

```bash
"could not delete test: remove /Users/sherif/.config/fabric/patterns/test: directory not empty"
```

since [`os.Remove`](https://pkg.go.dev/os#Remove) will only delete files or **empty** directories.

This PR replaces `os.Remove` with [`os.RemoveAll`](https://cs.opensource.google/go/go/+/go1.24.2:src/os/path.go;l=73) that deletes the specified path and any children it contains

This ensures that patterns can be deleted whilst also maintaining functionality of being able to delete entities such as contexts/sessions

### DIFF
--- a/plugins/db/fsdb/storage.go
+++ b/plugins/db/fsdb/storage.go
@@ -67,7 +67,7 @@ func (o *StorageEntity) GetNames() (ret []string, err error) {
 }
 
 func (o *StorageEntity) Delete(name string) (err error) {
-	if err = os.Remove(o.BuildFilePathByName(name)); err != nil {
+	if err = os.RemoveAll(o.BuildFilePathByName(name)); err != nil {
 		err = fmt.Errorf("could not delete %s: %v", name, err)
 	}
 	return


ksylvan (changes requested):
I don't like the proposed behavior.

If Fabric finds something in the patterns directory that it didn't expect, it must have been created and placed there by the user, so it seems right for Fabric to fail with a helpful error message.

The user can then decide what to do with what they placed in the patterns/ directory and try again.

sherif-fanous:
I understand the concern in case a user stores files unrelated to a pattern in a pattern directory.

With that said the current `DELETE` pattern API is technically of no use.

If the user will have to manually delete the files in the directory then why would they then use the API to delete the pattern?

I'd just do `rm -fr ~/.config/fabric/patterns/test` instead of going through the hassle of `rm ~/.config/fabric/patterns/test/*` then invoking the API

ksylvan:
> With that said the current `DELETE` pattern API is technically of no use.

Agreed.

PR #1463, opened 2025-05-07
Add completion to the build output for Nix

CodeCorrupt:
Add the completion scripts to the nix package build output so that completion works on supported shells.

## Related issues
None

## Screenshots
Build output puts the completion files in the output where nix expects them.
```
$ tree result/
result/
├── bin
│   ├── code_helper
│   ├── fabric
│   └── to_pdf
└── share
    ├── bash-completion
    │   └── completions
    │       └── fabric.bash
    ├── fish
    │   └── vendor_completions.d
    │       └── fabric.fish
    └── zsh
        └── site-functions
            └── _fabric
```
Screenshot shows tab completion working with Zsh after installing the package from my fork.
Screenshot 2025-05-07 at 5 22 50 PM

### DIFF
--- a/nix/pkgs/fabric/default.nix
+++ b/nix/pkgs/fabric/default.nix
@@ -2,6 +2,7 @@
   lib,
   buildGoApplication,
   go,
+  installShellFiles,
 }:
 
 buildGoApplication {
@@ -19,6 +20,13 @@ buildGoApplication {
   ];
 
   inherit go;
+
+  nativeBuildInputs = [ installShellFiles ];
+  postInstall = ''
+    installShellCompletion --zsh ./completions/_fabric
+    installShellCompletion --bash ./completions/fabric.bash
+    installShellCompletion --fish ./completions/fabric.fish
+  '';
 
   meta = with lib; {
     description = "Fabric is an open-source framework for augmenting humans using AI. It provides a modular framework for solving specific problems using a crowdsourced set of AI prompts that can be used anywhere";


ksylvan (approved):
I always compile from source, but this looks good to me.

ksylvan:
@CodeCorrupt Do you use Nix in your normal workflow? How do you like it?

My only real exposure has been that I  created docker-running functions to work with it when trying to debug some buld/test failures. For example, this runs the `nix flake check` that's run in GitHub Actions:

```bash
pushd ~/src/fabric

_green="$(tput setaf 2)"; _reset="$(tput sgr0)";
echo "${_green}"'Running "nix flake check" in the docker instance ...'"${_reset}"

docker run --rm -it -e HOME=/root -v $(pwd):/workdir -w /workdir nixos/nix bash -c 'git config --global --add safe.directory /workdir && mkdir -p ~/.config/nix && echo "experimental-features = nix-command flakes" > ~/.config/nix/nix.conf && git config --global init.defaultBranch main && nix flake check --all-systems && echo Running \"nix flake check\" done.'
echo ""

popd; unset _green _rese
```

ksylvan:
Personally, I'm pleased to see that more people are using the zsh, bash, and fish completions of Fabric!

Since putting it in a few weeks ago, there are a lot more instances of me using "fabric -p create_[TAB] -m grok[TAB]" instead of all the custom pattern aliases. 😄

ksylvan:
Whoah!!! I just figured out that using the aliases does not kill the command completion!

image

PR #1512, opened 2025-06-12
feat:Add support for Amazon Bedrock

### LINKED ISSUE #1500: [Feature request]: Add support for Amazon Bedrock
### What do you need?

Would you be open to a PR to add support for [Amazon Bedrock](https://aws.amazon.com/bedrock/)? 

I've had a brief look to understand the work involved:

* Create a new vendor in the `plugins/ai` directory (`bedrock`?)
* Create a client that fulfils the `Vendor` interface.
* Create associated tests for the new vendor.

Is there a place to document how to use each vendor? This would need updating as well.

Bedrock support was asked about in #1193. I opened a new issue as that was a question rather than a feature request.

ConnorKirk:
This PR adds support for using [Amazon Bedrock](https://aws.amazon.com/bedrock/) within fabric.

Users can select models provided by Bedrock for use within Fabric. 

### Models

I was unsure whether to hardcode the list of available models, or retrieving them from AWS. Both approaches have trade offs. Hardcoding the models is faster, as I couldn't see a good way to cache models (or easily clear the cache). It means that newer models will need to be added via PR, and some listed models might not be available for users.

It's possible to dynamically generate the list of available models using `ListInferenceProfiles` and `ListFoundationModels`. This could take a few seconds, so would be a poor experience for users unless it is cached. The advantage would be that the list of models returned would be specific to the user.

### Authorisation

AWS has a standard credential provider that the SDKs and CLI uses. I opted to use this, rather than prompting the user for the credentials with fabric. If a user already has credentials for AWS configured, these will be detected and used automatically.

[AWS Credential Provider](https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html)

### User Agent

The AWS SDK client has an extra key appended to the user agent. This helps track the impact of this work. I'm happy to remove it if preferred.

### Not included
* Only supports text content. Doesn't support multimodel content such as images.
* Doesn't support custom models
* Doesn't support Bedrock specific features like Guardrails or Prompt caching.

## Related issues

closes #1500


plugins/ai/bedrock/bedrock.go
```diff
@@ -0,0 +1,308 @@
...
+			Content: []types.ContentBlock{&types.ContentBlockMemberText{Value: msg.Content}},
+		}
+		messages = append(messages, message)
+
+	}
+
+	return
+}
+
+var MODELS = []string{
```
ksylvan, line 161: Is there a way to get the available models dynamically?

ksylvan (changes requested):

plugins/ai/bedrock/bedrock.go
ConnorKirk: Yes, it's possible if the user has AWS credentials configured. It's straight forward to implement.
I wasn't clear if it was called on each invocation of the tool, or if it was possible to cache. I kept it simple for the initial PR. Fetching dynamically would take a short period of time to complete.

ksylvan:
@ConnorKirk please look at PR #1514 - maybe even try to chat with the author - I'd like to see if the two of you can collaborate on the Bedrock plugin.

ConnorKirk:
Thanks @ksylvan. There are some good ideas in #1514 that could be incorporated

PR #1519, opened 2025-06-16
feat: Dynamically list AWS Bedrock models

ConnorKirk:
This PR updates the Bedrock plugin to fetch the available foundation models and inference providers.

The advantage of this approach is that the list of models is correct for the user. The downside is that it involves atleast two network requests/

## Related issues
This is a follow up PR for a change requested in #1512


ksylvan (approved):
Looks good @ConnorKirk thanks for the contribution

PR #1537, opened 2025-06-22
Added check for AWS_DEFAULT_REGION when AWS access keys are configured in the environment

andrewsjg:
Updated hasAWSCredentials to also check for AWS_DEFAULT_REGION when access keys are configured in the environment

Adds a check for AWS_DEFAULT_REGION in the hasAWSCredentials function. This is required when using AWS creds set in the environment. 

## Related issues
Extends: [1523](https://github.com/danielmiessler/fabric/pull/1523)

### DIFF
--- a/core/plugin_registry.go
+++ b/core/plugin_registry.go
@@ -44,7 +44,8 @@ import (
 func hasAWSCredentials() bool {
 	if os.Getenv("AWS_PROFILE") != "" ||
 		os.Getenv("AWS_ROLE_SESSION_NAME") != "" ||
-		(os.Getenv("AWS_ACCESS_KEY_ID") != "" && os.Getenv("AWS_SECRET_ACCESS_KEY") != "") {
+		(os.Getenv("AWS_ACCESS_KEY_ID") != "" && os.Getenv("AWS_SECRET_ACCESS_KEY") != "" && os.Getenv("AWS_DEFAULT_REGION") != "") {
+
 		return true
 	}
 


core/plugin_registry.go
```diff
@@ -44,7 +44,8 @@ import (
 func hasAWSCredentials() bool {
 	if os.Getenv("AWS_PROFILE") != "" ||
 		os.Getenv("AWS_ROLE_SESSION_NAME") != "" ||
-		(os.Getenv("AWS_ACCESS_KEY_ID") != "" && os.Getenv("AWS_SECRET_ACCESS_KEY") != "") {
+		(os.Getenv("AWS_ACCESS_KEY_ID") != "" && os.Getenv("AWS_SECRET_ACCESS_KEY") != "" && os.Getenv("AWS_DEFAULT_REGION") != "") {
```
ksylvan, line 47: This seems incorrect.

- AWS credentials (access key ID and secret access key) are authentication-related
- AWS region is a configuration parameter, not a credential
- Many AWS SDKs can function without `AWS_DEFAULT_REGION` by using service-specific endpoints or falling back to default regions

ksylvan (changes requested):

core/plugin_registry.go
andrewsjg: Fair, its not credential/auth related.  But I don't believe just the KEY_ID and ACCESS_KEY are enough to assume Bedrock is "Configured".  

When you go to configure your default model it throws the error:

"Bedrock operation error Bedrock: ListFoundationModels, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region"  If you only have the key and secret in the environment and no creds file/profile setup.

Perhaps there needs to be a check when the Bedrock controlplane client is created to ensure IsConfigured returns false if it can't resolve the region or service endpoint?

Apologies, I should have been more specific in my PR

andrewsjg: Maybe using AddSetupQuestion if the region cant be read would be a better solution?

andrewsjg: Created new pull request

PR #1565, opened 2025-07-01
Add OAuth auhentication support for Anthropic

ksylvan:
# Add OAuth auhentication support for Anthropic

## Summary

This PR implements OAuth authentication support for the Anthropic API client in Fabric, allowing users to authenticate using OAuth flow instead of API keys. Additionally, it includes improvements to environment variable naming conventions and updates to the REST API configuration endpoints to support the new authentication method.

### `plugins/ai/anthropic/anthropic.go`
Added OAuth authentication flow implementation for Anthropic API, including:
- OAuth configuration constants and endpoints
- PKCE (Proof Key for Code Exchange) implementation for secure OAuth flow
- Browser-based authorization flow
- Token management and storage

### `plugins/plugin.go`
Improved environment variable name sanitization by:
- Using regex to replace non-alphanumeric characters with underscores
- Trimming leading/trailing underscores for cleaner variable names

### `restapi/configuration.go`
Extended configuration endpoints to support the new Anthropic auth token:
- Added `anthropic_auth_token` field to configuration responses
- Added support for updating the auth token via REST API

### OAuth Implementation in `anthropic.go`

Added OAuth constants:
```go
const oauthClientID = "9d1c250a-e61b-44d9-88ed-5944d1962f5e"
const oauthAuthURL = "https://claude.ai/oauth/authorize"
const oauthTokenURL = "https://console.anthropic.com/v1/oauth/token"
const oauthRedirectURL = "https://console.anthropic.com/oauth/code/callback"
```

New OAuth flow function:
```go
func runOAuthFlow() (token string, err error) {
    // Generates PKCE verifier and challenge
    // Opens browser for user authorization
    // Exchanges authorization code for access token
}
```

Modified client configuration to support both API key and OAuth token:
```go
if an.AuthToken.Value != "" {
    opts = append(opts, option.WithAuthToken(an.AuthToken.Value))
} else {
    opts = append(opts, option.WithAPIKey(an.ApiKey.Value))
}
```

### Environment Variable Sanitization in `plugin.go`

Improved variable name building:
```go
re := regexp.MustCompile(`[^A-Z0-9]+`)
name = re.ReplaceAllString(name, "_")
name = strings.Trim(name, "_")
```

## Reason for Changes

1. **OAuth Support**: Many users prefer OAuth authentication over API keys for better security and easier credential management. This change provides an alternative authentication method for Anthropic API access.

2. **Environment Variable Naming**: The previous implementation only replaced spaces with underscores, which could lead to invalid environment variable names with special characters. The new regex-based approach ensures all non-alphanumeric characters are properly handled.

3. **REST API Updates**: The configuration endpoints needed to be updated to support the new authentication token field, allowing users to manage OAuth tokens through the REST API.

## Impact of Changes

- **Enhanced Security**: OAuth flow provides a more secure authentication method compared to static API keys
- **Better User Experience**: Users can authenticate through their browser without manually copying API keys
- **Backward Compatibility**: The implementation maintains full backward compatibility - users can still use API keys if they prefer
- **Cleaner Environment Variables**: The improved sanitization prevents potential issues with invalid environment variable names

## Test Plan

1. Test OAuth flow with valid Anthropic account credentials
2. Verify fallback to API key authentication when OAuth is not used
3. Test environment variable generation with various special characters
4. Verify REST API endpoints correctly handle the new auth token field
5. Ensure existing API key authentication continues to work

## Additional Notes

- The OAuth implementation uses PKCE for enhanced security, which is recommended for public clients
- The browser opening functionality supports multiple platforms (Linux, macOS, Windows)
- The auth token is stored as a setting and can be persisted across sessions
- Users can choose between OAuth and API key authentication based on their preference


ksylvan:
Stopped here: 

```text
fabric -p ai -m $MODEL_CLAUDE 'Why is the sky blue?'
POST "https://api.anthropic.com/v1/messages": 401 Unauthorized {"type":"error","error":{"type":"authentication_error","message":"OAuth authentication is currently not supported."}}
```

We have to figure out how Claude Code does its magic when it grabs the OAuth token.

eugeis:
LGTM

ksylvan:
Okay, @eugeis and @johnsaigle, I got it to work. I will be finalizing the PR soon.

We use the same technique that the `sst/opencode` repo uses to spoof being Claude Code.

Issue #1195, opened 2024-12-12
[Question]: Ollama Get "http://localhost:11434/api/tags": dial tcp 127.0.0.1:11434: connect: connection refused


guarapodecoco:
### What is your question?

Hi mates,  today i've installed fabric and im trying to run it powered by Ollama as default vendor but im having this issue with ollama api, i've configured it with its default port, i have to say that ollama is running in my local host ive checked. I was looking over and i found i may create the api extension on my local Ollama enviorment? ive tried localhost/api and there were no results in the navigator so this error may be from Ollama and not fabric? If anyone knows please help. Ive configured youtube api and the transcription tool is amazing!! great job

mrsatyp:
same here in windows .. but think it works via wsl linux in windows , need to confirm

cipolino85:
I've installed fabric with OpenAI  Installed also Ollama via Windows but I cannot configure Ollama as default vendor. Any ideas how to make it work for yt extract?

mrsatyp:
I think it might be related to setting the Ollama endpoint whatever machine or docker etc it's running on as 'listening' there is command for that in one of the guides 

Hopefully this resolves it if I'm on the same page as what you meant

🤞🏼

mrsatyp:
The rest on the fabric side will be normal as you would have entered your http://:port info 

Hopefully we can add multiple ollama endpoint at some point 
Ta

sepidre:
Hi there 
I got the same issue.
But I am running on docker.
One Ollama ROCM instance
One Open WebUI
One coolmast/fabric webinterface

The Open WebUI is running fine with the Ollama api but fabric is not.
No auth at all.

There is fabric v1.4.132 running the docker image.

Any ideas?

sepidre:
Same issue with v1.4.221

sepidre:
Just ignore it, had to use the container name ollama instead of localhost.

ksylvan:
In docker you can use the Ollama URL `host.docker.internal:

```text
Enter your Ollama URL (as a reminder, it is usually http://localhost:11434') (leave empty for 'http://localhost:11434' or type 'reset' to remove the value):
http://host.docker.internal:11434
```

To use an Ollama running on your docker host:

```text
~ $ ./fabric -L

Available models:

Ollama

        [1]     cogito:14b
        [2]     deepcoder:14b
        [3]     deepseek-r1:14b
        [4]     deepseek-r1:latest
        [5]     devstral:latest
[...]

PR #1645, opened 2025-07-22
[issue #1195] only call ollama and bedrock APIs when configured.

tonymet:
* Avoid Calling ollama and bedrock (aws) API when not configured

## Related issues
closes issue #1195 

# Steps to Repeat
1. set up fabric .env file ONLY with `GEMINI_API_KEY`
2.  run `fabric /listmodels`

### Before 

```
Ollama Get "http://localhost:11434/api/tags": dial tcp [::1]:11434: connectex: No connection could be made because the target machine actively refused it.
Bedrock failed to list foundation models: operation error Bedrock: ListFoundationModels, https response error StatusCode: 403, RequestID: fee42150-dcfa-4204-9ef0-f0e12d47fdd9, api error UnrecognizedClientException: The security token included in the request is invalid.

Available models:

Gemini
```

### AFTER

```
 go run .\cmd\fabric /listmodels

Available models:

Gemini

        [1]     aqa
        [2]     embedding-001
```

## OLD
* ~~`hasAwsCredentials` is broken .  Need to discuss cases where AWS credentials are available but not needed (e.g. AWS cli is used for other services, not fabric)~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -353,3 +353,6 @@ tmp/
 
 # Ignore .claude/
 .claude/
+
+# debug builds
+__debug*exe*
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,12 +6,20 @@
     "configurations": [
 
         {
-            "name": "Launch Package",
+            "name": "listmodels",
             "type": "go",
             "request": "launch",
             "mode": "auto",
             "program": "${workspaceFolder}\\cmd\\fabric",
             "args": ["/listmodels"]
+        },
+        {
+            "name": "basic request via gemini",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "program": "${workspaceFolder}\\cmd\\fabric",
+            "args": ["introduction to powershell"]
         }
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+
+        {
+            "name": "Launch Package",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "program": "${workspaceFolder}\\cmd\\fabric",
+            "args": ["/listmodels"]
+        }
+    ]
+}
--- a/internal/core/plugin_registry.go
+++ b/internal/core/plugin_registry.go
@@ -43,6 +43,7 @@ import (
 // potential authentication source exists so we can safely initialize the
 // Bedrock client without causing the AWS SDK to search for credentials.
 func hasAWSCredentials() bool {
+	return false
 	if os.Getenv("AWS_PROFILE") != "" ||
 		os.Getenv("AWS_ROLE_SESSION_NAME") != "" ||
 		(os.Getenv("AWS_ACCESS_KEY_ID") != "" && os.Getenv("AWS_SECRET_ACCESS_KEY") != "") {
@@ -62,6 +63,10 @@ func hasAWSCredentials() bool {
 		}
 	}
 	return false
+}
+
+func hasOllama() bool {
+	return os.Getenv("OLLAMA_URL") != ""
 }
 
 func NewPluginRegistry(db *fsdb.Db) (ret *PluginRegistry, err error) {
@@ -91,7 +96,6 @@ func NewPluginRegistry(db *fsdb.Db) (ret *PluginRegistry, err error) {
 	// Add non-OpenAI compatible clients
 	vendors = append(vendors,
 		openai.NewClient(),
-		ollama.NewClient(),
 		azure.NewClient(),
 		gemini.NewClient(),
 		anthropic.NewClient(),
@@ -100,6 +104,9 @@ func NewPluginRegistry(db *fsdb.Db) (ret *PluginRegistry, err error) {
 		perplexity.NewClient(), // Added Perplexity client
 	)
 
+	if hasOllama() {
+		vendors = append(vendors, ollama.NewClient())
+	}
 	if hasAWSCredentials() {
 		vendors = append(vendors, bedrock.NewClient())
 	}
--- a/internal/core/plugin_registry.go
+++ b/internal/core/plugin_registry.go
@@ -43,7 +43,10 @@ import (
 // potential authentication source exists so we can safely initialize the
 // Bedrock client without causing the AWS SDK to search for credentials.
 func hasAWSCredentials() bool {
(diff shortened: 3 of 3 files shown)


.vscode/launch.json
ksylvan, line 1: Please remove this change.

.gitignore
ksylvan, line 1: Please remove this change.

internal/core/plugin_registry.go
ksylvan, line 1: This is wrong. Bedrock will not activate unless you have credentials set up.

`hasAwsCredentials` in your patch always returns false which will break all Amazon Bedrock users.

internal/core/plugin_registry.go
```diff
@@ -64,6 +65,10 @@ func hasAWSCredentials() bool {
 	return false
 }
 
+func hasOllama() bool {
```
ksylvan, line 71: The Ollama provider also should not show up (or be hit) unless it's configured. Feels like there might be something wrong with your setup.

ksylvan (changes requested):
This PR should not even be necessary. Please re-check your setup.

ksylvan:
@tonymet Your PR would break Amazon Bedrock.

I'm pretty certain that both Bedrock and Ollama would not be hit unless they show up as "Configured"

image

I suggest you inspect your "fabric -S" setup and "reset" the OLLAMA provider (which will remove the Ollama URL and make the plugin inactive.

As for your Bedrock setup, unless you have AWS environment variables that make Fabric think you have a Bedrock setup available, it should not try to hit the Amazon Bedtock endpoint.

Let me know if you need help debugging and please close this PR.

tonymet:
thanks for the feedback

> I suggest you inspect your "fabric -S" setup and "reset" the OLLAMA provider (which will remove the Ollama URL and make the plugin inactive.

The BEFORE example is with a fresh installation.   You can see in the  Ollama.NewClient() always runs

> This is wrong. Bedrock will not activate unless you have credentials set up.

AWS credentials are used for AWS CLI .   I'd like advice on other Fabric config that could be used to prevent bedrock remote API calls

tonymet:
## Steps to Repeat
1. `cd $HOME\.config\`
2. `rm -rf fabric`
3. `fabric -S # choose Gemini , default model = 29`
4. `fabric /listmodels`

You can reproduce this using fabric /S + ollama reset.  Ollama remains configured in the .env file after "reset"

### Expected Results

ollama is not setup or called . Bedrock is not setup or called

### Actual Results

Ollama is self-configured and the call fails

```\
PS C:\Users\tonymet\.config> fabric /listmodels
Ollama Get "http://localhost:11434/api/tags": dial tcp [::1]:11434: connectex: No connection could be made because the target machine actively refused it.
Bedrock failed to list foundation models: operation error Bedrock: ListFoundationModels, https response error StatusCode: 403, RequestID: 75819941-87a8-4c38-b96b-dca4f109d66c, api error UnrecognizedClientException: The security token included in the request is invalid.
```

## Example Config
The following config is generated with only Gemini setup saved.  Note the OLLAMA settings.  The ollama config returns even when deleted
```
cat .\fabric\.env
DEFAULT_VENDOR=Gemini
DEFAULT_MODEL=gemini-2.5-flash
PATTERNS_LOADER_GIT_REPO_URL=https://github.com/danielmiessler/fabric.git
PATTERNS_LOADER_GIT_REPO_PATTERNS_FOLDER=data/patterns
PROMPT_STRATEGIES_GIT_REPO_URL=https://github.com/danielmiessler/fabric.git
PROMPT_STRATEGIES_GIT_REPO_STRATEGIES_FOLDER=data/strategies
BEDROCK_AWS_REGION=us-west-2
GEMINI_API_KEY=
OLLAMA_API_URL=http://localhost:11434
OLLAMA_HTTP_TIMEOUT=20m

```

tonymet:
Here's another example. 

## Steps
1. Fabric /S
2.  choose `[15]    Ollama (configured)`,  RESET
3. check for ollama config after saving

## Expected

no ollama env vars in .env .  Ollama does not show as `(configured)`

## ACTUAL
despite reset, ollama is configured 
```
fabric /S| select-string ollama

        [15]    Ollama (configured)
```

```
> cat .\fabric\.env |select-string ollama

OLLAMA_HTTP_TIMEOUT=20m

```

tonymet:
The common theme here are ollama and bedrock self-configuring and then calling the APIs.  ollama via `OLLAMA_HTTP_TIMEOUT` and bedrock via existing aws-cli credentials in ` $HOME\.aws`

ksylvan:
> The common theme here are ollama and bedrock self-configuring and then calling the APIs. ollama via `OLLAMA_HTTP_TIMEOUT` and bedrock via existing aws-cli credentials in ` $HOME\.aws`

Okay, Tony, I'll take care of that soon. Do you have any suggestions for the Bedrock AWS issue? (how do we tell if the aws credentials are just for AWS, versus Bedrock model usage?)

tonymet:
> > The common theme here are ollama and bedrock self-configuring and then calling the APIs. ollama via `OLLAMA_HTTP_TIMEOUT` and bedrock via existing aws-cli credentials in ` $HOME\.aws`
> 
> Okay, Tony, I'll take care of that soon. Do you have any suggestions for the Bedrock AWS issue? (how do we tell if the aws credentials are just for AWS, versus Bedrock model usage?)

only add BEDROCK_AWS_REGION to .env if that step is run in setup.   Then `hasAWSCredentials` can test for BEDROCK_AWS_REGION.  This would decouple aws-cli setup from Fabric setup

and it should not cause regressions because existing users would already have .env set up as expected.

ksylvan:
Thanks @tonymet - I'll be looking at this shortly. I really appreciate the investigation and the clear testcases here.

internal/core/plugin_registry.go
tonymet: i revised this to test for setup.  we will still need to fix `/setup` routine to avoid setting a default value for `BEDROCK_AWS_REGION` when not in use.

ksylvan:
I'll take care of this later tonight or tomorrow  thank you for the analysis.

I'll use a fresh docker container for the test case (since obviously my already configured home lab machines failed to surface these bugs)

ksylvan:
This is fixed in the latest Fabric. @tonymet please verify this works in your setup (especialy with AWS configured but no Bedrock).

tonymet:
yep it works great thanks!

I ran ..

```
fabric /setup
# configured default model
fabric /setup # again
# only my configurations are there as expected
 fabric /listmodels |select-string ollama
# ollama and bedrock are no longer called
```

great work thanks!

ksylvan:
You are welcome, and thanks for persisting, @tonymet, despite my being kind of a dick in my initial response. 😄

I love removing points of friction and making Fabric better and better, with the help of people like yourself.

tonymet:
you're too humble and no worries.  new version is great and a tad faster too

PR #1712, opened 2025-08-17
add pattern-model mapping with CLI management and automatic model selection

snafu4:
- Introduce helper functions to load, set, unset, and list pattern→model mappings in a user config file (pattern_models.yaml)
- Add a fabric pattern-model CLI subcommand supporting set, unset, and list operations for managing these mappings
- Automatically select vendor and model based on the mapping when running a pattern without an explicit model
- Document pattern-model mapping in a new guide and cross-link from the README
- Add tests verifying the correct config path for pattern-model files

Testing

⚠️ No automated tests were executed; review is based on static analysis only. Works for creator.

**Note**: I can barely spell golang.  All code entirely built by vibish coding.  First PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,6 +536,8 @@ cp completions/fabric.fish ~/.config/fish/completions/
 
 Once you have it all set up, here's how to use it.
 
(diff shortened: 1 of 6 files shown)


ksylvan (changes requested):
@snafu4 I appreciate your contribution. I'd like to think about it for a few days and see if we can come up with something much simpler.

ksylvan:
Implemented in a different way in #1714

PR #1713, opened 2025-08-17
docs: fix link to example Pattern in readme

kolaente:
I was trying to follow the link in the readme to the example Pattern only to get a 404. This change updates the link to the correct (I think) path.


ksylvan:
Thanks, @kolaente for this typo fix. Will roll it into the next feature I'm about to release (OpenAI speech-to-text transcription support).

PR #1770, opened 2025-09-14
Use Garble to get around false positive Microsoft Defender virus detection

ksylvan:
# Use Garble to get around false positive Microsoft Defender virus detection

## Summary

This PR integrates `garble` into the build pipeline to obfuscate the compiled Go binaries during release. Garble is a Go code obfuscation tool that makes reverse engineering more difficult by scrambling literals and reducing binary size.

## Files Changed

- **`.github/workflows/release.yml`**: Added a step to install the `garble` tool before running GoReleaser in the CI/CD pipeline.
- **`.goreleaser.yaml`**: Modified the build configuration to use `garble` as the Go compiler with obfuscation flags enabled.
- **`.vscode/settings.json`**: Added new words to the VS Code spell checker dictionary related to the garble integration.

### `.github/workflows/release.yml`
```yaml
+ - name: Install garble
+   run: go install mvdan.cc/garble@latest
```
Added installation of garble tool using `go install` to fetch the latest version from the official repository.

### `.goreleaser.yaml`
```yaml
+ gobinary: garble
+ # From https://github.com/eyevanovich/garble-goreleaser-example/blob/main/.goreleaser.yaml
+ # command is a single string.
+ # garble's 'build' needs the -literals and -tiny args before it, so we
+ # trick goreleaser into using -literals as command, and pass -tiny and
+ # build as flags.
+ command: "-literals"
+ flags: [ "-tiny", "-seed=random", "build" ]
```
Configured GoReleaser to use `garble` instead of the standard Go compiler. The configuration uses a workaround to properly pass garble's flags:
- `-literals`: Obfuscates string literals in the code
- `-tiny`: Further reduces binary size by removing additional debug information
- `-seed=random`: Ensures different obfuscation patterns across builds

### `.vscode/settings.json`
```json
+ "gobinary",
+ "mvdan",
```
Added dictionary entries for spell checking support in VS Code.

## Reason for Changes

The implementation of garble obfuscation serves to:
1. **Protect intellectual property**: Makes it significantly harder to reverse engineer the compiled binaries
2. **Reduce binary size**: The `-tiny` flag strips additional debug information, resulting in smaller release artifacts
3. **Enhance security**: Obfuscated string literals make it harder to extract sensitive information or understand the program's logic through static analysis

## Impact of Changes

- **Binary size**: Release binaries will be smaller due to the `-tiny` flag
- **Performance**: Minimal runtime impact; obfuscation primarily affects binary structure, not execution speed
- **Debugging**: Production binaries will be harder to debug, but this only affects release builds, not development builds
- **Build time**: Slightly increased build time during releases due to the obfuscation process
- **Reproducibility**: Each build will have different obfuscation patterns due to `-seed=random`, which enhances security but means binaries won't be byte-for-byte reproducible

## Test Plan

1. Verify the GitHub Actions workflow successfully installs garble
2. Confirm GoReleaser builds complete without errors
3. Test that the resulting binaries function correctly on all target platforms (Linux, macOS, Windows)
4. Validate that binary sizes are reduced compared to non-obfuscated builds
5. Attempt basic reverse engineering tools on the binary to confirm obfuscation is effective

## Additional Notes

- This is an attempt o implement the fix in [this Reddit thread](https://www.reddit.com/r/golang/comments/s1bh01/goexecutables_and_windows_defender/?share_id=TziSPnMDbktxAkl1fPxtz&utm_content=1&utm_medium=ios_app&utm_name=ioscss&utm_source=share&utm_term=1) about a way to get around the frequent false positive hits from Microwoft Defender on Windows that is holding up the Winget releases.
- The configuration follows a documented workaround pattern from the garble-goreleaser example repository
- This change only affects release builds; local development builds will continue to use the standard Go compiler
- Future consideration: Monitor garble updates for potential breaking changes or new obfuscation features


ksylvan:
This PR was completely replaced by 1771.

PR #1914, opened 2026-01-01
feat(code_helper): add stdin support for piping file lists

### LINKED ISSUE #1869: [Feature request]: make code_helper use its standard input
### What do you need?

FWIU, `code_helper` is good at preparing a JSON input containing a code base to send to an AI.

Currently, it takes everything it finds in a given directory but some files depending on the ignore option (or the lack of it, its default being ".git,node_modules,vendor").
In order for this tool to be closer to the UNIX philosophy, I would have asked it to take the list of files to consider from its standard input.

Doing this would make scenarii like these ones possible:

```console
git ls-files | code_helper | fabric ...
find . -name '*.py' | code_helper | fabric ...
find . | grep -E '.*.pth' | code_helper | fabric ...
```

It would also avoid the burden of supporting other options like `--depth` and all other features that users might need in the future.
All these needs would then just be a matter of providing the good input to `code_helper` by delegating it to other CLI commands.

Aside from this, I would perhaps rename `code_helper` to somewhat closer to its purpose, perplexity proposed these names:

* fs2json
* file2json
* file-json

I am pretty sure there are even better names reflecting the purpose of `code_helper` but I must admit I am not good at finding sound names.

**NOTE**: While reading the description of `code_helper`, I found myself comparing it to `tar` but instead of producing a binary archive, it delivers a "JSON archive".

majiayu000:
Closes #1869

## Changes
- Add stdin support to code_helper for piping file lists from commands like `find` or `git ls-files`
- Add ScanFiles() function to process explicit file lists
- Update usage text with stdin examples
- Add tests for the new functionality

### DIFF
--- a/cmd/generate_changelog/incoming/1914.txt
+++ b/cmd/generate_changelog/incoming/1914.txt
@@ -1,0 +1,5 @@
+### PR [#1914](https://github.com/danielmiessler/Fabric/pull/1914) by [majiayu000](https://github.com/majiayu000): feat(code_helper): add stdin support for piping file lists
+
+- Added stdin support for piping file lists to code_helper, enabling commands like `find . -name '*.go' | code_helper "instructions"` and `git ls-files '*.py' | code_helper "Add type hints"`
+- Implemented automatic detection of stdin pipe mode with single argument (instructions) support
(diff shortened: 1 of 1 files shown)


ksylvan (approved):
Nicely done @majiayu000 - Clean diff and nice separation of concerns.

PR #1982, opened 2026-02-03
Added an integration script to create opencode prompts from fabric patterns

mrchainman:
The title says it.
This script creates commands under ~/.config/opencode/commands/ from the fabric patterns.
I am not sure if this is the right place for this script...
Feedback is welcome.


ksylvan:
@mrchainman The idea (and the shell script implementation) is simple enough that I can't justify adding it to the repo as an official integration. You are welcome to create a discussion post in the "Show and Tell" category to let other users know how they can do something similar.

Thank you for the contribution.

PR #2006, opened 2026-02-16
feat: When running from a symlink, use the executable name as the pattern argument

konstantint:

Provides an alternative way of setting up aliases via symlinks:

```
ln -s /path/to/fabric ~/.local/bin/summarize
cat stuff.txt | summarize
summarize -u <stuff>
...
```

Namely, when the binary's name is not 'fabric' and the -p flag is not provided, it will assume the binary's name is the pattern. This makes it easier to manage aliases for a few select patterns without needing to edit `.bashrc` (there may be situations where this is inconvenient for some reasons).

Seems like an easy and neat feature to have here, right?

### DIFF
--- a/cmd/generate_changelog/incoming/2006.txt
+++ b/cmd/generate_changelog/incoming/2006.txt
@@ -1,0 +1,3 @@
+### PR [#2006](https://github.com/danielmiessler/Fabric/pull/2006) by [konstantint](https://github.com/konstantint): feat: When running from a symlink, use the executable name as the pattern argument
+
+- Feat: When running from a symlink, use the executable name as the pattern argument


ksylvan:
Hi @konstantint. I appreciate that the change is small and focused, but it's a bad idea. We have 250+ patterns, so you could possibly have a couple of hundred symlinks pointing to your fabric binary. In addition, many of the patterns have generic names like `summarize` that are too close, in my opinion, to commands like `sum`.

What does this do that you can't already achieve with the bash/zsh aliases that are documented in the README?

For example I have all my patterns aliased (with an `f_` prefix) so I can do things like

```
f_summarize_git_diff -m $MODEL_CLAUDE
```

konstantint:
I'm not sure I understand your counterarguments here.

> it's a bad idea.

In what sense do you find this functionality hurts the project? It surely does not prevent one from using aliases yet offers an alternative for situations where aliases are not a suitable option. 5 LOC doesn't look like much of a complexity creep either, right?

> We have 250+ patterns, so you could possibly have a couple of hundred symlinks pointing to your fabric binary.

Firstly, so what? Secondly, I personally need just three of those 250, for example.

> In addition, many of the patterns have generic names like summarize that are too close, in my opinion, to commands like sum.

But it is up to each user to decide whether they prefer summarize or f_summarize, right? (I bet most would pick the former).

> What does this do that you can't already achieve with the bash/zsh aliases that are documented in the README?

As I mentioned, there are occasional situations where aliases can be less convenient than symlinks. For example:

1. If you use more than one shell you need to maintain them in multiple places. If you version your dotfiles, maintenance of these aliases adds even more complexity.
2. Some people may prefer to maintain a very tight list of commands and find it is easier to type ln -s / rm to add/remove them rather than edit their dotfiles.
3. If you need to execute the commands from elsewhere (non-shell scripts, Makefiles) you cannot use aliases.
4. It is sometimes easier to mount a directory with symlinks into a docker container rather than modify dotfiles of whatever user lives there.
5. It may be convenient to provide a united set of tools to multiple users of a system by just putting a few symlinks into /usr/bin rather than forcing aliases in everyone's profile.

ksylvan:
Okay. I'm willing to rethink my stance.

ksylvan:
On second thought, @konstantint , after playing with it, I kind of ike the idea. Thanks.

PR #2047, opened 2026-03-06
Explicitly instruct YT summary task to be aware of possible transcription errors.

bvandevliet:
Self-explanatory ;)


data/patterns/youtube_summary/system.md
```diff
@@ -19,21 +20,13 @@ Take a step back and think step-by-step about how to achieve the best possible r
 ## OUTPUT INSTRUCTIONS
 
 - Only output Markdown
-
```
ksylvan, line 22: Why are you removing these instructions?

bvandevliet: I'm only removing the empty new lines in between them.

ksylvan: I misread it on my phone. Now I see. :-) Thanks.

PR #2059, opened 2026-03-13
Claude code Integration Feature and Integration Tests

adamwrose:

Add an integration to claude code. 
- I was introduced to your project from NetworkChuck and became facinated with the work of fabric. I currently have an github copilot and claude (Anthropic pro) subscription.  I noticed that the anthropic API was charged differently and claude code could use the Subscription vs paying $5 per x many tokens.  I hope this will make it to where this is useful for more users with a tighter budget. If you like this, I can make one for opencode as well. 

- prerequisites
  - install claude code from either npm or brew. 
  - have an anthropic subscription.
## Related issues

Please reference any open issues this PR relates to in here.
If it closes an issue, type `closes #[ISSUE_NUMBER]`.

## Screenshots

I have written a lot of tests to validate my work and done some manual testing as well. 

_Great Work_,
Adam


ksylvan:
@adamwrose Did Claude write this? What is the problem this is solving? How would I run it and test it?

ksylvan:
@adamwrose Your PR description has a piece of a conversation (with some large language model?) in it....

> I have written a lot of tests to validate my work and done some manual testing as well.
> 
> Great Work,
> Adam

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil
+	}
+
+	args := c.buildArgs(opts, system)
+	args = append(args, "--verbose", "--output-format", "stream-json", "--include-partial-messages")
+	args = append(args, userPrompt)
+
+	binary := c.getBinary()
+	debuglog.Debug(debuglog.Detailed, "ClaudeCode SendStream launching: %s %s\n", binary, strings.Join(args, " "))
+	cmd := exec.Command(binary, args...)
```
ksylvan: **[Major]** `SendStream` uses `exec.Command` but `Send` uses `exec.CommandContext`. If the `claude` subprocess hangs or the caller disconnects, this process cannot be killed — it will orphan.

**Suggestion:** Use `exec.CommandContext` with an internal `context.WithCancel` or `context.WithTimeout`, consistent with `Send()` at line 183.

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+		args = append(args, "--add-dir", filepath.Dir(opts.ImageFile))
+	}
+	return args
+}
+
+func (c *Client) Send(ctx context.Context, msgs []*chat.ChatCompletionMessage, opts *domain.ChatOptions) (string, error) {
+	system, userPrompt := c.extractMessages(msgs, opts)
+	if userPrompt == "" {
+		return "", nil
+	}
```
ksylvan, line 184: **[Major]** When `userPrompt` is empty after message extraction, `Send` returns `("", nil)` — a successful result with no content. Callers cannot distinguish "model returned nothing" from "no input was provided."

**Suggestion:** Return an error instead:
```go
return "", fmt.Errorf("claude: no prompt content after message extraction")
```

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+		"claude-sonnet-4-5-20250929",
+		"claude-opus-4-1-20250805",
+		"claude-sonnet-4-20250514",
+		"claude-sonnet-4-0",
+		"claude-4-sonnet-20250514",
+		"claude-opus-4-0",
+		"claude-opus-4-20250514",
+		"claude-4-opus-20250514",
+	}, nil
+}
```
ksylvan: **[Suggestion]** This static model list will become outdated as Anthropic releases new models. Consider querying `claude --list-models` dynamically (if the CLI supports it), falling back to this static list on error.

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+	case domain.ThinkingMedium:
+		args = append(args, "--effort", "medium")
+	case domain.ThinkingHigh:
+		args = append(args, "--effort", "high")
+	}
+	if opts.ImageFile != "" {
+		args = append(args, "--add-dir", filepath.Dir(opts.ImageFile))
+	}
+	return args
+}
```
ksylvan, line 178: **[Minor]** `ChatOptions` fields like `Temperature`, `TopP`, `MaxTokens`, and `Seed` are silently dropped. Consider passing `--max-tokens` if the CLI supports it, and logging a debug warning for unsupported options so users aren't confused when their settings have no effect.

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+		text, ok := parseStreamDelta(line)
+		if !ok {
+			continue
+		}
+		channel <- domain.StreamUpdate{
+			Type:    domain.StreamTypeContent,
+			Content: text,
+		}
+	}
```
ksylvan: **[Minor]** If the channel consumer stops reading, this send blocks the goroutine forever. A `select` with a `ctx.Done()` channel would enable graceful shutdown (this ties into the `exec.CommandContext` suggestion above).

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+	return strings.TrimSpace(stdout.String()), nil
+}
+
+func (c *Client) SendStream(msgs []*chat.ChatCompletionMessage, opts *domain.ChatOptions, channel chan domain.StreamUpdate) error {
+	defer close(channel)
+
+	system, userPrompt := c.extractMessages(msgs, opts)
+	if userPrompt == "" {
+		return nil
+	}
```
ksylvan, line 219: **[Minor]** Same silent no-op as `Send` — returns `nil` (success) when there's no prompt. Should return an error for consistency.

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+// subprocess uses local Claude Code session auth and does not detect nested
+// Claude Code execution.
+func cleanEnv() []string {
+	env := os.Environ()
+	filtered := env[:0:0]
+	for _, e := range env {
+		if !strings.HasPrefix(e, "CLAUDECODE=") && !strings.HasPrefix(e, "ANTHROPIC_API_KEY=") {
+			filtered = append(filtered, e)
+		}
+	}
```
ksylvan, line 327: **[Minor]** Only strips `CLAUDECODE` and `ANTHROPIC_API_KEY`. Other `ANTHROPIC_*` env vars (e.g. `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`) pass through. Consider using a `strings.HasPrefix(e, "ANTHROPIC_")` check instead.

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
...
+func (c *Client) Send(ctx context.Context, msgs []*chat.ChatCompletionMessage, opts *domain.ChatOptions) (string, error) {
+	system, userPrompt := c.extractMessages(msgs, opts)
+	if userPrompt == "" {
+		return "", nil
+	}
+
+	args := c.buildArgs(opts, system)
+	args = append(args, userPrompt)
+	binary := c.getBinary()
+	debuglog.Debug(debuglog.Detailed, "ClaudeCode Send launching: %s %s\n", binary, strings.Join(args, " "))
```
ksylvan: **[Minor]** Full command args (including the user prompt) are logged at debug level. For long prompts this could be noisy or leak sensitive content into logs. Consider truncating the prompt portion.

internal/plugins/ai/claudecode/claudecode.go
```diff
@@ -0,0 +1,293 @@
+package claudecode
```
ksylvan, line 3: **[Suggestion]** Missing package-level doc comment. A brief `// Package claudecode ...` would help with `go doc` and IDE hover.

internal/plugins/ai/claudecode/claudecode.go
ksylvan, line 260: **[Suggestion]** The `message` event from Claude's stream-json output contains token usage metadata (`input_tokens`, `output_tokens`). Currently this is discarded. Forwarding it as a `StreamTypeUsage` update would give callers visibility into costs.

internal/cli/binary_test_helpers_test.go
```diff
@@ -0,0 +1,243 @@
...
+		entry.err = cmd.Run()
+		entry.log = stderr.String()
+	})
+
+	if entry.err != nil {
+		t.Fatalf("build %s: %v\n%s", relativeDir, entry.err, entry.log)
+	}
+
+	return entry.path
+}
```
ksylvan, line 175: **[Suggestion]** `buildGoBinary` compiles Go binaries at test time. While the `sync.Once` caching is smart, this still adds significant CI wall time. Consider a `TestMain` that pre-builds all needed binaries, or a Makefile target so CI can cache them between runs.

internal/cli/binary_integration_test.go
```diff
@@ -0,0 +1,227 @@
...
+		t.Fatalf("write Claude wrapper: %v", err)
+	}
+	if err := os.Chmod(wrapperPath, 0o700); err != nil {
+		t.Fatalf("chmod Claude wrapper: %v", err)
+	}
+
+	return []string{"CLAUDECODE_BINARY_PATH=" + wrapperPath}
+}
+
+func TestIntegrationClaudeCodePatternExecution(t *testing.T) {
```
ksylvan, line 49: **[Minor]** Integration tests don't set per-test timeouts. If a Claude API call hangs, the entire CI job blocks until the global timeout. Consider adding `t.Deadline()` checks or explicit `context.WithTimeout` for each test (e.g. 2-5 minutes).

adamwrose:
I have not forgotten about this.  I have been doing research on this for the last couple of weeks. There are opensource projects that expose claude code as a service that fabric and openclaw can talk to.  What do you think.  Want me to clean this up where AI helped me on the test cases and commit or do you want me to document a way to use claude code with an open source API?

ksylvan:
@adamwrose Thanks for your response. I'd love for us to be able to use `claude` in this way without running afoul of the Anthropic Terms of Service. As far as I understand it, those claude code proxies are a gray area.

Tell me what you're thinking, let's discuss!

ksylvan (changes requested):

adamwrose:
> @adamwrose Did Claude write this? What is the problem this is solving? How would I run it and test it?

claude helped with the testing.

internal/plugins/ai/claudecode/claudecode.go
adamwrose: this has been fixed.

internal/plugins/ai/claudecode/claudecode.go
adamwrose: done.

internal/plugins/ai/claudecode/claudecode.go
adamwrose: there is not a claude --list-models .  I see a claude --model where you can specify a model but that is the best I could find.

internal/plugins/ai/claudecode/claudecode.go
adamwrose: Added logUnsupportedOptions() — emits debuglog.Debug warning for each non-zero unsupported field

internal/plugins/ai/claudecode/claudecode.go
adamwrose: Wrapped in select { case channel <- update: case <-ctx.Done(): return ctx.Err() }

internal/plugins/ai/claudecode/claudecode.go
adamwrose: Added truncateArgs() helper that caps the last arg (the prompt) at 200 chars

internal/plugins/ai/claudecode/claudecode.go
adamwrose: Added package doc comment

internal/plugins/ai/claudecode/claudecode.go
adamwrose: Added Usage field to streamEvent struct; parseStreamEvent now returns StreamTypeUsage updates for message events

internal/cli/binary_test_helpers_test.go
adamwrose: working on it.

ksylvan:
@adamwrose What is the status of this branch?

internal/cli/binary_integration_test.go
adamwrose: I have fixed this

adamwrose:
> @adamwrose What is the status of this branch?

Sorry for such a delay.  my day job issues have pulled me away but I was able to spend some time on this. 
I committed the following changes: 
-  Stream cancellation now shuts down the ClaudeCode subprocess cleanly, integration test timeouts now
-  use context cancellation instead of panic, and I added coverage for the cancellation path.
-  I also reran the ClaudeCode integration test and go test ./..., both passing.

internal/cli/binary_test_helpers_test.go
adamwrose: fixed

ksylvan:
Thanks for this, @adamwrose - Superceded by #2226

PR #2069, opened 2026-03-19
feat(cli): add --workflow for sequential pattern composition

1rashiid:
Adds a `--workflow <file>` flag that runs multiple patterns in sequence, piping each step's output into the next step's input. Workflows are simple YAML/JSON files with a `steps:` list — each step names a pattern and may optionally override `input`, `variables`, `model`, or `vendor`.

Every step reuses the existing `Chatter.Send` path, so all current flags (`--stream`, `--dry-run`, `-m`, `-v`, `--context`, etc.) keep working. Patterns are validated against the local store **before** any LLM call, and runtime failures stop immediately with a `[step N/TOTAL pattern] failed: …` error. Progress goes to stderr; final result to stdout. Single-pattern `-p` usage is untouched.

```bash
cat article.txt | fabric --workflow wf.yaml --stream
```


ksylvan:
Please fix the `modernize` errors, @1rashiid - Thanks.,

PR #2091, opened 2026-04-11
Update Dockerfile for best practices and critical CVE fixes

jimscard:
## Summary
- update `scripts/docker/Dockerfile` to align the container build with current industry best practices
- pin Alpine 3.21 and Go 1.25.9 explicitly and install Go directly in the builder stage
- fix multiple critical CVEs, including the `setuptools` fix for CVE-2025-47273, and run the final image as a non-root user

## Testing
- docker build -t fabric-pr-check -f scripts/docker/Dockerfile .

### DIFF
--- a/scripts/docker/Dockerfile
+++ b/scripts/docker/Dockerfile
@@ -1,11 +1,28 @@
 # syntax=docker/dockerfile:1
 
-FROM golang:1.25-alpine AS builder
+ARG GO_VERSION=1.25.9
+ARG ALPINE_VERSION=3.21
+
+FROM alpine:${ALPINE_VERSION} AS builder
+
+ARG GO_VERSION
+ARG TARGETARCH
+ENV GOROOT=/usr/local/go
+ENV PATH=${GOROOT}/bin:${PATH}
 
 WORKDIR /src
 
-# Install build dependencies
-RUN apk add --no-cache git
+# Install build dependencies and the pinned Go toolchain.
+RUN apk add --no-cache ca-certificates curl git tar \
+    && case "${TARGETARCH:-$(apk --print-arch)}" in \
+        amd64|x86_64) go_arch='amd64' ;; \
+        arm64|aarch64) go_arch='arm64' ;; \
+        *) echo "Unsupported TARGETARCH: ${TARGETARCH:-$(apk --print-arch)}" >&2; exit 1 ;; \
+    esac \
+    && curl -fsSL "https://go.dev/dl/go${GO_VERSION}.linux-${go_arch}.tar.gz" -o /tmp/go.tgz \
+    && rm -rf "${GOROOT}" \
+    && tar -C /usr/local -xzf /tmp/go.tgz \
+    && rm -f /tmp/go.tgz
 
 COPY go.mod go.sum ./
 RUN go mod download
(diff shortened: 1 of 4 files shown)


scripts/docker/Dockerfile
```diff
@@ -14,15 +31,22 @@ COPY . .
...
+    && mkdir -p /root/.config/fabric \
+    && apk upgrade --no-cache
 
 COPY --from=builder /fabric /usr/local/bin/fabric
 
 EXPOSE 8080
 
+# Run as non-root user
+RUN adduser -D appuser
+USER appuser
```
ksylvan: **Bug: Config directory mismatch with runtime user**

`mkdir -p /root/.config/fabric` creates the config dir under `/root/`, but the container runs as `appuser` (home = `/home/appuser/`). `fabric` resolves config via `os.UserHomeDir()` → `/home/appuser`, so the pre-created directory is never found, and `appuser` can't write to `/root/`. Fabric will fail at runtime trying to load or create its config.

**Suggested fix:**
```dockerfile
RUN adduser -D appuser \
    && mkdir -p /home/appuser/.config/fabric \
    && chown -R appuser:appuser /home/appuser/.config
USER appuser
```

scripts/docker/Dockerfile
```diff
@@ -1,11 +1,28 @@
...
-# Install build dependencies
-RUN apk add --no-cache git
+# Install build dependencies and the pinned Go toolchain.
+RUN apk add --no-cache ca-certificates curl git tar \
+    && case "${TARGETARCH:-$(apk --print-arch)}" in \
+        amd64|x86_64) go_arch='amd64' ;; \
+        arm64|aarch64) go_arch='arm64' ;; \
+        *) echo "Unsupported TARGETARCH: ${TARGETARCH:-$(apk --print-arch)}" >&2; exit 1 ;; \
+    esac \
+    && curl -fsSL "https://go.dev/dl/go${GO_VERSION}.linux-${go_arch}.tar.gz" -o /tmp/go.tgz \
```
ksylvan, line 26: **Security: No checksum verification on Go tarball**

The Go toolchain is downloaded over HTTPS but not verified against a known SHA256 hash before extraction. A compromised or tampered download could inject a malicious toolchain into the build.

**Suggested fix** — add SHA256 verification after download:
```dockerfile
ARG GO_SHA256_AMD64="<sha256-hash>"
ARG GO_SHA256_ARM64="<sha256-hash>"
# ...after curl download:
    && echo "${expected_sha256}  /tmp/go.tgz" | sha256sum -c - \
```

ksylvan (changes requested):
I fixed an unrelated *modernize* GoLang bug that was causing tests to fail on your branch.

I have a couple of changes I'd like to make.

Thank you, @jimscard

scripts/docker/Dockerfile
ksylvan: Fixed in [`56a1a7f2`](https://github.com/danielmiessler/fabric/commit/56a1a7f2). Config directory now created under `/home/appuser/.config/fabric` with proper ownership via `chown -R appuser:appuser`.

scripts/docker/Dockerfile
ksylvan: Fixed in [`56a1a7f2`](https://github.com/danielmiessler/fabric/commit/56a1a7f2). Added `GO_SHA256_AMD64`/`GO_SHA256_ARM64` build ARGs with official checksums from `go.dev`, verified via `sha256sum -c` before extraction.

ksylvan (approved):

PR #2092, opened 2026-04-11
feat(openai): add GrokAI search grounding via xAI Responses API

Resistor52:

Fixes GrokAI web search grounding. Today, `fabric -V GrokAI -m grok-4-fast-reasoning --search "query"` fails with HTTP 422 Unprocessable Entity because `internal/plugins/ai/openai/openai.go` hardcodes OpenAI's `web_search_preview` tool name in `buildResponseParams`. xAI's Responses API expects `web_search`, and additionally accepts an `x_search` live grounding tool. This PR plumbs per-provider overrides so GrokAI can emit the correct tool types while every other provider retains current behavior.

## Bug reproduction

Before this change:

```
$ fabric -V GrokAI -m grok-4-fast-reasoning --search "What is the current status of the James Webb Space Telescope?"
HTTP 422 Unprocessable Entity
```

## Fix design

Two new fields are added to `openai_compatible.ProviderConfig`:

- `WebSearchToolName string` (default empty): when non-empty, overrides the default `web_search_preview` tool type string emitted on the Responses API.
- `EnableXSearch bool` (default false): when true, appends an additional `{"type":"x_search"}` tool entry alongside the web search tool.

These fields are plumbed through `openai_compatible.NewClient` into new unexported fields on `openai.Client` via two small setters (`SetWebSearchToolName`, `SetEnableXSearch`). `buildResponseParams` consults them when `opts.Search` is true:

1. If `webSearchToolName != ""`, the existing `responses.ToolParamOfWebSearchPreview` helper is invoked with the override string cast to `responses.WebSearchToolType`. The SDK marshals the `type` field verbatim, so the emitted JSON is `{"type":"web_search"}` for xAI.
2. If `enableXSearch == true`, a second `ToolUnionParam` is appended carrying `{"type":"x_search"}`. The SDK's `WebSearchToolParam` is reused as a minimal container because every non-type field is `omitzero` and elided during marshalling.

The GrokAI entry in `ProviderMap` now sets `WebSearchToolName: "web_search"` and `EnableXSearch: true`. The `--search` flag description in `internal/cli/flags.go` is updated to list Grok alongside Anthropic, OpenAI, and Gemini.

### Go SDK approach

This is approach A from a short investigation of `github.com/openai/openai-go@v1.12.0`: cast a plain string to `responses.WebSearchToolType` and rely on the fact that `WebSearchToolParam.Type` marshals the value verbatim without SDK-side enum validation. No raw JSON, no `SetExtraFields` hack, no new dependencies, and the change stays entirely within the typed SDK surface. Alternative approaches (custom `ToolUnionParam`, raw extra-field JSON tools array, `map[string]any` tools) were considered and rejected as unnecessary once it was confirmed that the SDK's marshaller passes the string through unmodified.

## Backwards compatibility

Every existing provider leaves both new fields at their zero values (`""` and `false`). The default code path in `buildResponseParams` still produces exactly one `web_search_preview` tool entry, byte for byte identical to current behavior. A dedicated regression test (`TestBuildResponseParams_DefaultProvider_Unchanged`) asserts this. The existing `TestBuildResponseParams_WithSearch` and `TestBuildResponseParams_WithSearchAndLocation` tests still pass unchanged.

## Test coverage

Three new tests in `internal/plugins/ai/openai/openai_test.go`:

- `TestBuildResponseParams_GrokAI_WithSearch`: constructs a client with `SetWebSearchToolName("web_search")` and `SetEnableXSearch(true)`, runs `buildResponseParams` with `Search: true`, and asserts that `params.Tools` has length 2 with type strings `web_search` and `x_search` in that order.
- `TestBuildResponseParams_DefaultProvider_Unchanged`: default `NewClient()` with `Search: true` still emits exactly one tool entry whose type is `web_search_preview`.
- `TestBuildResponseParams_GrokAI_WithoutSearch`: GrokAI-configured client with `Search: false` emits nil `params.Tools`, confirming the no-search path does not regress.

Full suite results:

```
$ go test ./...
ok  	github.com/danielmiessler/fabric/internal/plugins/ai/openai	5.460s
ok  	github.com/danielmiessler/fabric/internal/plugins/ai/openai_compatible	0.003s
(all other packages PASS)
```

`go build ./...`, `gofmt -l .`, and `go vet ./...` all clean.

## Live verification

With a real `GROKAI_API_KEY`, after this change:

```
$ fabric -V GrokAI -m grok-4-fast-reasoning --search "What is the current operational status of the James Webb Space Telescope? One sentence with a source URL."

As of March 2026, the James Webb Space Telescope remains in active mission status, continuing to conduct groundbreaking observations of the universe from its orbit at the Sun-Earth L2 Lagrange point without any reported issues.[[1]](https://science.nasa.gov/mission/webb)

## Sources

- [1](https://science.nasa.gov/mission/webb)
```

Real content, real NASA URL returned by the live xAI Responses API. No more HTTP 422.

## Related issues

None.

## Screenshots

N/A, backend change.


ksylvan (approved):
Thanks @Resistor52 - Approved.

PR #2101, opened 2026-04-17
fix(youtube): preserve ffmpeg frame timing in visual OCR cues

shaun0927:
## Summary

This PR fixes the YouTube visual OCR path so its cue timestamps are derived from ffmpeg frame timing instead of the OCR result index.

## Problem

`GrabVisual()` currently extracts frames using either `fps=%d` or `select='gt(scene,%f)'`, but later formats subtitle-like cues from:

```go
secs := i
```

So the nth OCR result is labeled as second `n` regardless of when that frame actually occurred in the source video.

## Fix

- append `showinfo` to the ffmpeg filter used for visual extraction
- parse `pts_time` values from ffmpeg output
- format cue ranges from actual frame timestamps (falling back cleanly if timestamps are unavailable)
- add regression coverage for the timestamp parser, cue-range helper, and end-to-end visual extraction path

## Files changed

- `internal/tools/youtube/youtube.go`
- `internal/tools/youtube/timestamp_test.go`
- `internal/tools/youtube/visual_timestamp_test.go`

## Tests

```bash
go test ./internal/tools/youtube -v
```

## Related issues

Closes #2099

### DIFF
--- a/cmd/generate_changelog/incoming/2101.txt
+++ b/cmd/generate_changelog/incoming/2101.txt
@@ -1,0 +1,17 @@
+### PR [#2101](https://github.com/danielmiessler/Fabric/pull/2101) by [shaun0927](https://github.com/shaun0927): fix(youtube): preserve ffmpeg frame timing in visual OCR cues
+
+- Preserve real visual OCR timing in YouTube frame cues
+The visual OCR path emitted subtitle-style cues using the OCR result index
+instead of the underlying ffmpeg frame time, which made the output look more
+precise than it was. This change parses ffmpeg showinfo timestamps, formats
+cue ranges from real frame times, and adds regression coverage for the timing
+helpers and the end-to-end visual extraction path.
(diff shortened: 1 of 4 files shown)


cmd/generate_changelog/incoming/2101.txt
```diff
@@ -0,0 +1,17 @@
+### PR [#2101](https://github.com/danielmiessler/Fabric/pull/2101) by [shaun0927](https://github.com/shaun0927): fix(youtube): preserve ffmpeg frame timing in visual OCR cues
```
ksylvan, line 1: The changelog snippet ends up in the ChangeLog and on the Release Notes. It needs to be short, concise, but also accurate. It shouldn't contain all the details or be a mini-PRD. Pleaese make this short and brief.
You can also use the `--ai-summarize` flag of the `generate_changelog` CLI tool.

ksylvan (changes requested):

PR #2102, opened 2026-04-17
fix(chat): emit SSE error and skip complete on prompt setup failures

### LINKED ISSUE #2100: /chat SSE returns complete even when strategy loading fails
## Description
When the REST `/chat` endpoint receives a prompt with an invalid `strategyName`, the request fails during prompt/session construction, but the SSE response still ends with a final `complete` event instead of an `error` event.

That makes clients interpret the request as successful even though no valid response was produced.

## Affected area
- `internal/server/chat.go`
- current upstream I validated locally: `9743e10273c45e1c377c10ec0ffaf0de8fefb3c8` (`v1.4.448`)

## Minimal reproduction
Send a `/chat` request with a missing strategy name, for example:

```json
{
  "prompts": [
    {
      "userInput": "hello",
      "vendor": "TestVendor",
      "model": "test-model",
      "strategyName": "missing-strategy"
    }
  ]
}
```

## Current behavior
The handler logs an internal error from `chatter.Send()` but still closes the stream normally and emits:

```json
{"type":"complete","format":"plain","content":""}
```

with no SSE `error` event.

## Expected behavior
If request assembly fails before a usable response stream is produced, the SSE stream should emit an `error` event and should not emit a final `complete` event for that failed prompt.

## Why this matters
For SSE clients, `complete` strongly implies success. In this case the request actually failed, so the endpoint violates its own streaming contract and can cause silent client-side misclassification of failures.

## Local validation evidence
I reproduced this locally with a focused regression test:

```bash
go test ./internal/server -run 'TestHandleChat_InvalidStrategyEmitsErrorAndSkipsComplete' -v
```

I have a narrow fix prepared with a regression test if maintainers are open to a PR.

shaun0927:
## Summary

This PR fixes a REST `/chat` SSE contract bug where prompt/session setup failures could still end with a final `complete` event.

## Problem

When `chatter.Send()` fails before any stream-originated error update is emitted (for example, when `strategy.LoadStrategy()` fails during request assembly), the handler currently:

1. logs the error,
2. closes the stream channel, and then
3. emits a final `complete` event.

That means an SSE client can interpret the request as successful even though no valid response was produced.

## Fix

- forward an explicit `StreamTypeError` update when `chatter.Send()` returns an error that was not already surfaced through stream updates
- suppress the final `complete` event whenever an error was emitted for that prompt
- add regression coverage for the invalid-strategy path

## Files changed

- `internal/server/chat.go`
- `internal/server/chat_test.go`

## Tests

```bash
go test ./internal/server -v
```

## Related issues

Closes #2100

### DIFF
--- a/cmd/generate_changelog/incoming/2102.txt
+++ b/cmd/generate_changelog/incoming/2102.txt
@@ -1,0 +1,17 @@
+### PR [#2102](https://github.com/danielmiessler/Fabric/pull/2102) by [shaun0927](https://github.com/shaun0927): fix(chat): emit SSE error and skip complete on prompt setup failures
+
+- Prevent false SSE completion after prompt setup failures
+The /chat handler previously relied on stream-originated error updates and
+still emitted a terminal complete event when chatter setup failed before any
+stream updates were sent. This change forwards a server-side error event for
+those early failures, suppresses complete after errors, and adds regression
+coverage for an invalid strategy path.
+Constraint: Keep the fix narrow to the REST SSE path
+Rejected: Broad chatter API redesign | unnecessary for a targeted contract fix
+Confidence: high
+Scope-risk: narrow
+Reversibility: clean
+Directive: Preserve the invariant that complete is emitted only for successful prompt runs
+Tested: go test ./internal/server -v
+Not-tested: End-to-end browser SSE client behavior outside package tests
+Related: #2100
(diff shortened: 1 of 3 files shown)


cmd/generate_changelog/incoming/2102.txt
```diff
@@ -0,0 +1,17 @@
+### PR [#2102](https://github.com/danielmiessler/Fabric/pull/2102) by [shaun0927](https://github.com/shaun0927): fix(chat): emit SSE error and skip complete on prompt setup failures
+
```
ksylvan, line 2: Hi @shaun0927 while I appreciate that you included this snippet, I'll give the same feedback I gave in your other PR:

The changelog snippet ends up in the ChangeLog and on the Release Notes. It needs to be short, concise, but also accurate. It shouldn't contain all the details or be a mini-PRD. Pleaese make this short and brief.
You can also use the `--ai-summarize` flag of the `generate_changelog` CLI tool.

ksylvan (changes requested):

PR #2114, opened 2026-05-09
feat: shell portability fixes and bun lockfile

Skippy-the-Magnificent-one:
## Changes

- **Shebangs:** `#!/bin/bash` → `#!/usr/bin/env bash` across shell scripts for portability (nix, macOS, non-standard bash paths)
  - `scripts/installer/install.sh`
  - `internal/plugins/template/Examples/remote-security-report.sh`
  - `internal/plugins/template/Examples/security-report.sh`
  - `internal/plugins/template/Examples/track_packages.sh`
- **web:** Added `bun.lock` for bun users (bun doesn't use pnpm-lock.yaml)
- **web:** Removed `pnpm-lock.yaml` (redundant with `package-lock.json` already in repo)
- **docs:** Removed empty `docs/voices/README.md`

## Why

The `#!/bin/bash` shebang assumes bash lives at `/bin/bash`, which isn't true on NixOS, some BSDs, or custom installs. `#!/usr/bin/env bash` is the portable standard.

The web directory now has both `package-lock.json` (npm) and `pnpm-lock.yaml` -- having a `bun.lock` as well supports all three package managers. Since `pnpm-lock.yaml` is redundant with the existing npm lockfile, this swaps it for bun support.


ksylvan:
Hi @Skippy-the-Magnificent-one — thanks for this, and for the clear write-up. The shebang portability point is spot-on: `#!/usr/bin/env bash` is the right call for NixOS/macOS/non-standard installs, and we'd genuinely like to land that fix.

Unfortunately I can't merge this PR as it stands, because it bundles three separable decisions into one change:

1. **Shebang fixes** (4 scripts) — this part is great and mergeable on its own.
2. **Swapping `pnpm-lock.yaml` for `bun.lock`** — which package managers the web app officially supports is a project-level decision. It affects CI, contributor docs, and every web contributor's workflow, so it needs its own issue/discussion rather than landing as a side effect of a shell portability PR. (You're right that having both `package-lock.json` and `pnpm-lock.yaml` in `web/` is untidy — worth resolving deliberately, one way or the other.)
3. **Deleting `docs/voices/README.md`** — small correction: that file isn't empty. It documents the six Gemini TTS voice samples and the commands to regenerate them, so we'd like to keep it.

The fastest path forward: if you open a fresh PR with just the shebang changes, I'd expect it to merge quickly. And if bun support matters to you, please do open an issue for it — happy to have that conversation on its own merits.

Thanks again for contributing, and sorry to be the bearer of a "please split it up" — we know it's extra work, but it keeps each decision reviewable and reversible. 🙏

PR #2117, opened 2026-05-16
fix: clearer error when binary name used as pattern fallback

### LINKED ISSUE #2112: [Bug]: Misleading "pattern '<binary-name>' not found" when no pattern is specified, plus README zsh yt() example is broken since #1960
### What happened?

`fabric-ai` falls back to using its own binary name as the pattern name when called without `-p` but with positional args or with `-y URL --transcript`, producing this confusing error:

```
could not get pattern fabric-ai: pattern 'fabric-ai' not found.
Run 'fabric -l' to see available patterns
```

This appears to be a side effect of #1960 ("consume all positional arguments as input"). It hits two common entrypoints:

**1. Stale README zsh `yt()` helper.** The current shell-integration snippet in the README defines:

```bash
yt() {
    ...
    fabric -y "$video_link" $transcript_flag
}
```

With v1.4.452, `yt <url>` produces the binary-name-not-a-pattern error. Adding `-p extract_wisdom` (or any other pattern) makes it work again.

**2. Anyone trying `fabric-ai update` (or any positional-arg "subcommand-ish" invocation), e.g. expecting a CLI subcommand pattern.** Same misleading error — easy to misread as "fabric-ai isn't installed" or "Homebrew packaging is broken" before realizing it's the CLI looking up its own binary name as a pattern.

### Repro

```bash
# v1.4.452, installed via Homebrew (homebrew-core formula)

fabric-ai update
# → could not get pattern fabric-ai: pattern 'fabric-ai' not found.

fabric-ai -y "https://www.youtube.com/watch?v=b2znvI2ZBd8" --transcript
# → could not get pattern fabric-ai: pattern 'fabric-ai' not found.

fabric-ai -y "https://www.youtube.com/watch?v=b2znvI2ZBd8" --transcript -p extract_wisdom
# ✓ works as expected
```

### Expected

Either (a) a clear error like *"no pattern specified — use -p PATTERN"* and a non-zero exit, or (b) print-only behaviour for `--transcript` when no pattern is set (matches the wording in `--help`: *"...send to chat **or** print it..."*).

In any case, the fallback to using `os.Args[0]` as a pattern name is a bug — the binary name will never be a valid pattern, and the resulting error message sends users hunting in the wrong direction.

### Also: README update

The shell-integration `yt()` example in the README should be updated to either:
- include a default pattern (e.g. `-p extract_wisdom`), or
- pipe the raw transcript out (and let users compose with `| fabric-ai -p ...`).

### Version

- fabric-ai 1.4.452 (homebrew-core formula `fabric-ai`, Apple Silicon)
- macOS 26.4

OdinKral:
## Problem

Closes #2112.

When `fabric` is invoked via a symlink or renamed binary (e.g. `fabric-ai`) **without `-p`**, the binary name is silently used as the pattern name. If that pattern doesn't exist, the user sees:

```
could not get pattern fabric-ai: pattern 'fabric-ai' not found. Run 'fabric -l' to see available patterns
```

There is no hint that the binary name was used as the fallback, so the user doesn't know to add `-p`.

## Fix

Adds a `patternFromBinaryName()` helper in `flags.go` that checks whether the current pattern matches the binary-name fallback condition (mirrors the existing fallback logic). When `handleChatProcessing` returns an error and this condition is true, the error is wrapped with a hint:

```
Hint: no -p flag was given; binary name 'fabric-ai' was used as the pattern name. Use -p <pattern> to specify a pattern explicitly.
```

The symlink use-case (intentionally naming a binary after a pattern) is **not broken** — it still works the same way; the hint only appears when the pattern lookup fails.

## Changes

- `internal/cli/flags.go` — add `patternFromBinaryName()` helper
- `internal/cli/cli.go` — wrap chat error with hint when binary-name fallback is the likely cause
- `internal/i18n/locales/en.json` — add `pattern_from_binary_name_hint` i18n key
- `internal/cli/flags_test.go` — 6 unit tests for `patternFromBinaryName()`

## Test plan

- [ ] Run `go test ./internal/cli/...` — new `TestPatternFromBinaryName` tests pass
- [ ] Symlink `fabric` to `fabric-ai`, run `echo "hi" | fabric-ai` without patterns installed — verify new hint appears in error output
- [ ] Run with `-p summarize` as normal — verify no change in behavior

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -110,7 +110,11 @@ func Cli(version string) (err error) {
 	}
 
 	// Handle chat processing
-	err = handleChatProcessing(currentFlags, registry, messageTools)
+	if err = handleChatProcessing(currentFlags, registry, messageTools); err != nil {
+		if patternFromBinaryName(currentFlags.Pattern) {
+			err = fmt.Errorf("%w\n%s", err, fmt.Sprintf(i18n.T("pattern_from_binary_name_hint"), currentFlags.Pattern))
+		}
+	}
 	return
 }
 
--- a/internal/cli/flags.go
+++ b/internal/cli/flags.go
@@ -556,6 +556,14 @@ func (o *Flags) IsChatRequest() (ret bool) {
 	return
 }
 
+// patternFromBinaryName returns true when the pattern matches the binary name fallback
+// (i.e., no -p flag was given and the binary was not named "fabric", "main", or "cmd").
+func patternFromBinaryName(patternName string) bool {
+	execName := filepath.Base(os.Args[0])
+	execName = strings.TrimSuffix(execName, filepath.Ext(execName))
+	return execName != "fabric" && execName != "main" && execName != "cmd" && execName != "" && patternName == execName
+}
+
 func (o *Flags) WriteOutput(message string) (err error) {
 	fmt.Println(message)
 	if o.Output != "" {
--- a/internal/cli/flags_test.go
+++ b/internal/cli/flags_test.go
@@ -482,3 +482,30 @@ func TestExtractFlag(t *testing.T) {
 		})
 	}
 }
+
+func TestPatternFromBinaryName(t *testing.T) {
+	oldArgs := os.Args
+	defer func() { os.Args = oldArgs }()
+
+	tests := []struct {
+		name        string
+		binaryName  string
+		patternName string
+		expected    bool
+	}{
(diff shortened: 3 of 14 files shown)


OdinKral:
Superseded by #2173 — same commits, resubmitted from a personal fork (`OdinKral/fabric-contrib`) rather than the `AnsibleBBridge` org fork. It has also been rebased onto current main — this one had gone `CONFLICTING` after #2167 touched the same locale files.

No review changes — please continue at #2173.

PR #2118, opened 2026-05-16
fix(completions): add --list-transcription-models and --show-metadata to zsh completion

OdinKral:
## Summary

Two flags added in recent releases were absent from `completions/_fabric`:

| Flag | Added in |
|------|----------|
| `--list-transcription-models` | transcription feature |
| `--show-metadata` | `--show-metadata` flag |

The `--transcribe-model` completion already calls `_fabric_transcription_models()` (the dynamic list helper), but the standalone `--list-transcription-models` list flag was not in the completion spec at all. `--show-metadata` was also missing.

## Changes

- `completions/_fabric` — two lines added in the correct position next to their related flags

## Test plan

- [ ] Source the updated `_fabric` completion and verify `fabric --list<Tab>` shows `--list-transcription-models`
- [ ] Verify `fabric --show<Tab>` shows `--show-metadata`


OdinKral:
Superseded by #2174 — same commits, resubmitted from a personal fork (`OdinKral/fabric-contrib`) rather than the `AnsibleBBridge` org fork.

No review changes — please continue at #2174.

PR #2121, opened 2026-05-16
fix(completions): add --list-transcription-models and --show-metadata to bash and fish

OdinKral:
## Summary

Follow-up to #2118 (zsh). The same two flags were also missing from `fabric.bash` and `fabric.fish`:

- `--list-transcription-models` — absent from bash opts string and fish
- `--show-metadata` — absent from bash opts string and fish

Keeps all three completion scripts in sync with the current flag set.

## Changes

- `completions/fabric.bash` — added both flags to the opts string
- `completions/fabric.fish` — added two `complete` lines next to related flags


OdinKral:
Superseded by #2177 — same commits, resubmitted from a personal fork (`OdinKral/fabric-contrib`) rather than the `AnsibleBBridge` org fork.

No review changes — please continue at #2177.

PR #2123, opened 2026-05-16
fix: block path traversal in pattern name lookup

### LINKED ISSUE #2094: Harden pattern name input against path traversal in getFromDB
## Summary

The `getFromDB` method in `internal/plugins/db/fsdb/patterns.go` (line ~121) joins the user-supplied `name` parameter directly into file paths via `filepath.Join` without explicit path traversal sanitization. While the risk is **low**, adding a guard would be a good defensive hardening measure.

## Details

In `getFromDB`, the pattern name flows into two `filepath.Join` calls:

```go
customPatternPath := filepath.Join(o.CustomPatternsDir, name, o.SystemPatternFile)
// ...
patternPath := filepath.Join(o.Dir, name, o.SystemPatternFile)
```

A malicious or accidental `name` value like `../../etc` could theoretically traverse outside the intended patterns directory. In practice, the risk is very low because:

1. `filepath.Join` normalizes paths (removes redundant separators, resolves `.` elements)
2. The resulting path must contain a valid `system.md` file to succeed
3. Fabric is a local CLI tool, so the user already has filesystem access

However, as defense-in-depth, a simple guard would prevent any future risk if this code path is ever exposed through a network-facing interface.

## Suggested Fix

Add a path traversal check at the top of `getFromDB`, for example:

```go
if strings.Contains(name, "..") {
    return nil, fmt.Errorf("invalid pattern name: %q", name)
}
```

Or use `filepath.Rel` to verify the resolved path stays within the intended directory.

## Context

This was identified during the review of PR #2093. It is a **pre-existing** issue, not introduced by that PR. Filed separately as recommended in the review.

**Priority**: Low
**Category**: Security hardening

OdinKral:
## Summary

- `getFromDB` joined user-supplied pattern names directly into `filepath.Join` with no sanitisation, allowing names like `../etc/passwd` to escape the patterns directory
- Added a `strings.Contains(name, "..")` guard at the top of `getFromDB` that returns an error immediately
- Added `pattern_invalid_name` i18n key to all 11 locale files (English + 10 translations)
- Added `TestGetFromDB_PathTraversal` test covering 5 traversal variants

Closes #2094.

## Test plan

- [ ] `go test ./internal/plugins/db/fsdb/...` passes with new `TestGetFromDB_PathTraversal` test
- [ ] Normal pattern names (no `..`) continue to work
- [ ] Names with `..` anywhere return an error containing "invalid pattern name"


ksylvan:
Closing this PR unmerged — we could not merge it due to the way it was constructed (we are unable to push maintainer fixes to your fork branch):

```
git push --force
ERROR: Permission to AnsibleBBridge/Fabric.git denied to ksylvan.
fatal: Could not read from remote repository.
```

The change itself is good and will be re-submitted with you credited as author.

ksylvan:
Thank you @OdinKral for finding and fixing this path traversal vulnerability (and for the thorough test coverage) — the change itself was great, and your commit is preserved with your authorship.

Re-submitted with the needed maintainer fixes (i18n translations, changelog entry) as #2167: https://github.com/danielmiessler/Fabric/pull/2167

PR #2131, opened 2026-05-28
feat(vendor): add llama.cpp vendor plugin

xb17:
## What

Adds a dedicated vendor plugin for [llama.cpp](https://github.com/ggml-org/llama.cpp) server.

## Background

Issue #2072 requested llama.cpp support. Currently users have to route through the `openai_compatible` driver pointed at `http://localhost:8080/v1`, which works but requires manual setup and doesn't expose llama.cpp-specific behaviour.

## Why a dedicated driver

llama.cpp's server is OpenAI-compatible but differs in a few meaningful ways:

| | `openai_compatible` | `llama.cpp` |
|---|---|---|
| Default URL | varies | `http://localhost:8080/v1` |
| API key | required field | optional (local server) |
| `cache_prompt` | not sent | ✓ sent — reuses KV cache for repeated system prompts |
| SDK dependency | OpenAI Go SDK | none (hand-rolled HTTP) |

The `cache_prompt: true` field is particularly valuable for Fabric usage patterns, where the same system prompt (pattern) is sent repeatedly across requests. llama.cpp reuses the KV cache for the matching prefix, significantly reducing time-to-first-token on subsequent requests.

## Implementation

Follows the same hand-rolled HTTP approach as the LM Studio plugin. Implements `ListModels`, `SendStream`, and `Send`. API key is optional — the Authorization header is only set when a key is configured.

New files:
- `internal/plugins/ai/llamacpp/llamacpp.go`
- i18n keys in `internal/i18n/locales/en.json`
- Registration in `internal/core/plugin_registry.go`

Closes #2072


gschoenberger:
Any chance to get this change merged? I am a user of llama.cpp and would like set up fabric with it 👍

Issue #2108, opened 2026-05-01
Groq chat.completions rejects Fabric payload when pattern input becomes a system
 message (last role must be 'user')


das-analyst:
Title: Groq chat.completions rejects Fabric payload when pattern input becomes a system message (last role must be 'user')

Summary

Fabric can produce a chat payload containing only a system message (pattern instructions concatenated with the INPUT). Groq’s chat.completions API rejects this with: "last message role must be 'user'". That breaks many built-in patterns (e.g., summarize) when using the Groq vendor.

Environment
- Fabric CLI v1.4.451
- Groq vendor configured via ~/.config/fabric/.env (GROQ_API_BASE_URL=https://api.groq.com/openai/v1, DEFAULT_MODEL=groq/compound)

Reproduction
1. Ensure Groq config present in ~/.config/fabric/.env.
2. Run:

   printf 'Test input for summary.' | fabric -p summarize --vendor Groq --model groq/compound --debug=4

3. Observe error and debug lines showing a single system message and the Groq response:

   POST "https://api.groq.com/openai/v1/chat/completions": 400 Bad Request {"message":"last message role must be 'user'"}

Expected
When sending to chat.completions, the final message in the messages array should be role='user'. Fabric should either:
- Send patterns as separate messages (system + user) so the final message role is 'user', or
- Detect that Groq requires a user message and append a user message (e.g., with the input) before sending.

Suggested fix (minimal)
- In the Groq/OpenAI-compatible vendor adapter, when constructing the messages array:
  - If messages currently end with role != 'user' (or there is only a system message), append a user message with the input content (or move the input into a final user message) before POST.
- Add unit/integration test covering a pattern that places the INPUT placeholder in system.md to ensure Groq chat requests always end with a user role.

Optional notes
- Many patterns place the INPUT marker in system.md by design; the adapter should be robust to that by ensuring the last message role is 'user' for chat endpoints.

I can provide debug logs and the summarize pattern files on request.

genevera:
This happens also in vLLM.  #2137 fixes this

PR #2137, opened 2026-06-07
middleware: add NormalizeInputShape to handle system-only message arrays

genevera:

Some LLM backends (vLLM, certain Bedrock endpoints) reject requests that contain only system-role messages. NormalizeInputShape promotes the last system message to user when no user message is present, which transparently handles the case where a pattern absorbs {{input}} into its system message.

- domain.NormalizeInputShape is idempotent (no-op when user present)
- Called at the session boundary before vendor.Send()
- Makes the DeepSeek system-to-user fallback redundant
- 8 test cases covering all edge cases


genevera:
fixes #2108

ksylvan:
Why is this PR touching so much? What coding tools generated it @genevera ?

genevera:
@ksylvan It was Opencode.. I have omniroute set up with a number of combos (random choice between a random opencode-zen model, a random nvidia nim model, a local instance of Qwen3.5-122B-A10B. 

It touches so much because I wanted to resolve the issues brought up during code reviews.  I thought about squashing all the commits but didn’t know if yall are ok with force pushing or not.

i know it’s a lot.  Should i close this and give it another go?

PR #2140, opened 2026-06-15
fix: include Grok in --search help text, completions, and locales

Resistor52:
## What

`--search` supports Grok (xAI) web search since #2092, but `fabric --help` still shows:

```
--search    Enable web search tool for supported models (Anthropic, OpenAI, Gemini)
```

The capability works (`-V GrokAI -m grok-4.3 --search` grounds correctly), but the help text, shell completions, and README still omit Grok, so the support is undiscoverable.

## Root cause

`--help` renders flag descriptions from the i18n message catalog, not the go-flags struct tag. `internal/cli/help.go` maps the flag to a key:

```go
"search": "enable_web_search_tool",
```

#2092 updated the struct tag in `internal/cli/flags.go` (`...Gemini, Grok)`) but not the catalog value, so the rendered help kept the old vendor list.

## Change

Append `Grok` to the `enable_web_search_tool` value in all 11 locales (matching each locale's list separator, including the full-width `、` for `ja`/`zh` and the Arabic comma for `fa`), the zsh and fish completions, and the generated README help block. No functional change; 14 files, one line each.

## Test plan

- `go build ./cmd/fabric && ./fabric --help | grep -- --search` now shows `(Anthropic, OpenAI, Gemini, Grok)`.
- `go test ./internal/i18n/...` passes.
- All locale JSON files validate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,7 +719,7 @@ Application Options:
       --liststrategies              List all strategies
       --listvendors                 List all vendors
       --shell-complete-list         Output raw list without headers/formatting (for shell completion)
-      --search                      Enable web search tool for supported models (Anthropic, OpenAI, Gemini)
+      --search                      Enable web search tool for supported models (Anthropic, OpenAI, Gemini, Grok)
       --search-location=            Set location for web search results (e.g., 'America/Los_Angeles')
       --image-file=                 Save generated image to specified file path (e.g., 'output.png')
       --image-size=                 Image dimensions: 1024x1024, 1536x1024, 1024x1536, auto (default: auto)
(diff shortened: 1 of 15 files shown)


ksylvan:
Thanks for this, @Resistor52 — and especially for the writeup. Your
diagnosis was exactly right: `--help` renders flag descriptions from the
i18n message catalog (`internal/cli/help.go` maps `search` →
`enable_web_search_tool`), so updating the struct tag in #2092 alone left
Grok undiscoverable. That detail is what made the fix obvious.

The changes have all landed, just split across two PRs:

- #2179 — the zsh and fish completion strings
- #2180 — the `enable_web_search_tool` value in all 11 locales, plus the
  generated README help block

For #2180 I applied your locale hunks directly from this PR's patch, so
your translations went in byte-for-byte, including the locale-specific
list separators (`، ` for fa, `、` for ja/zh). You're credited as
co-author on 97308081.

I verified nothing was dropped by reverse-applying this PR's diff against
`main` — all 14 files apply clean, meaning every line here is now present
upstream. So I'm closing this as fully merged rather than as declined.

Thanks again!

PR #2163, opened 2026-07-16
Add Requesty as an OpenAI-compatible provider

Thibaultjaigu:
## Add Requesty as an OpenAI-compatible provider

This adds [Requesty](https://requesty.ai) as an OpenAI-compatible AI provider by mirroring the existing OpenRouter entry.

Requesty is an OpenAI-compatible LLM router that exposes 400+ models behind a single endpoint using `provider/model` naming, so it fits the existing `openai_compatible` provider mechanism directly.

### Changes
- `internal/plugins/ai/openai_compatible/providers_config.go`: add a `Requesty` entry to `ProviderMap`, mirroring `OpenRouter`: name `Requesty`, base URL `https://router.requesty.ai/v1`. The provider is auto-registered by the existing registry loop, and the `REQUESTY_API_KEY` env var is derived from the provider name by the existing plugin machinery (same as `OPENROUTER_API_KEY`).
- `README.md` / `README.zh.md`: list Requesty in the OpenAI-compatible providers section.

### Testing
- `go build ./...`: clean
- `go vet ./internal/plugins/ai/openai_compatible/`: clean
- `gofmt -l` on the changed file, no output
- `go test ./internal/plugins/ai/openai_compatible/`: ok
- Verified live against the Requesty endpoint (same base URL + Bearer auth): a call to `openai/gpt-4o-mini` returned the expected completion with HTTP 200.

Note: `internal/server/configuration.go` (web-UI config endpoint) only lists a subset of the compatible providers, so I left it untouched to keep the diff minimal and matching the actual OpenRouter wiring, happy to add the web-UI key field too if you'd like.

Config: get a key at https://app.requesty.ai/api-keys, model list at https://app.requesty.ai/router/list, docs at https://docs.requesty.ai.

---

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,6 +374,7 @@ Fabric supports a wide range of AI providers:
 - Mistral
 - Novita AI
 - OpenRouter
+- Requesty
 - SiliconCloud
 - Together
 - Venice AI
--- a/README.md
+++ b/README.md
@@ -375,6 +375,7 @@ Fabric supports a wide range of AI providers:
 - Novita AI
 - OpenRouter
 - Pzero
+- Requesty
 - SiliconCloud
 - Synthorai
 - Together
--- a/README.zh.md
+++ b/README.zh.md
@@ -358,7 +358,7 @@ fabric --setup
 
 **OpenAI 兼容供应商：**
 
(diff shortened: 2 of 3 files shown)


Thibaultjaigu:
Rebased onto current main to clear the README.zh.md conflict. No functional changes, build and tests pass.

PR #2126, opened 2026-05-16
improve: expand compare_and_contrast and create_flash_cards; apply cognitive science to flashcards

OdinKral:
## Summary

Two stub patterns expanded with real structure. `create_flash_cards` additionally updated with cognitive science principles from active recall research.

### `compare_and_contrast`
**Before:** "Please be brief. Compare and contrast the list of items." with no dimension-selection guidance.

**After:** Steps for picking meaningful dimensions (not trivial similarities), cell-length guidance, and a **Key Takeaways** section to surface non-obvious patterns.

### `create_flash_cards`
**Before:** 4 bullets, no card count, no format spec.

**After (initial):** Numbered steps, card count target, Q:/A: format.

**After (cognitive science pass):** Added principles from active recall research:

- **Retrieval vs recognition**: Questions must force the learner to *generate* the answer, not pattern-match the phrasing. Wrong/right example included in the pattern itself.
- **One idea per card**: Explicitly enforced — compound questions get split.
- **Concrete anchors**: Abstract answers must end with a concrete example `(e.g. ...)`.
- **Three question types**: Definition, Mechanism/Why-How, Application/When — all used where the material supports it.
- **Synthesis cards**: 2–4 per deck connecting concepts into chunks, building usable mental models.
- **Card ordering**: Definition → Mechanism → Application → Synthesis mirrors how understanding builds.

`raw_query` was also flagged by the audit script but intentionally left as-is — it's a minimal pass-through by design.

## Test plan

- [ ] `fabric -p compare_and_contrast "Python vs JavaScript vs Go"` returns a table with meaningful dimensions and a Key Takeaways section
- [ ] `fabric -p create_flash_cards` on a short article returns Q:/A: cards across all three question types, with concrete examples on abstract answers, and 2–4 synthesis cards at the end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2125, opened 2026-05-16
fix: add # INPUT section to 93 patterns that were missing it

OdinKral:
## Summary

93 patterns were missing a `# INPUT` section — the standard marker Fabric uses to delimit where user input is injected into a pattern.

Three variants were found and fixed:

| Variant | Count | Fix |
|---------|-------|-----|
| `## INPUT` (wrong heading level) | 7 | Upgraded to `# INPUT` |
| Bare `INPUT:` text with no heading | 8 | Added `# INPUT` header before it |
| No input section at all | 78 | Appended `# INPUT\n\nINPUT:` |

All changes are **structural only** — no pattern content, instructions, or output format was altered. This is a pure normalization of the input boundary marker.

## Why it matters

- The `# INPUT` heading is what `--readpattern` and pattern tooling use to display where user content goes
- Inconsistent markers make it harder to programmatically parse patterns
- New contributors following the pattern template expect this section to exist

## Test plan

- [ ] `fabric -p analyze_bill "..."` still works (pattern that had no INPUT section)
- [ ] `fabric -p translate "..."` still works (pattern that had `## INPUT`)  
- [ ] `fabric -p coding_master "..."` still works (pattern that had bare `INPUT:`)
- [ ] `grep -rL "^# INPUT" data/patterns/*/system.md` returns empty (all patterns now have the section)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2124, opened 2026-05-16
fix: strip embedded content from bloated patterns; add maintenance audit script

OdinKral:
## Summary

This PR removes large embedded content payloads from patterns (saving ~430KB / ~100K tokens per call across the affected patterns), fixes shell completion gaps, and adds a maintenance audit script.

### Pattern fixes

| Pattern | Before | After | Saved |
|---------|--------|-------|-------|
| `improve_prompt` | 518 lines — 490 from verbatim OpenAI/GPT-4 guide | 45 lines, model-agnostic principles | ~10KB |
| `explain_math` | 8-line stub | Full IDENTITY + STEPS + 5 output sections | — |
| `generate_code_rules` | 8-line stub | STEPS + OUTPUT + merge discipline | — |
| `extract_insights_dm` | 226KB — full podcast transcript as EXAMPLE | Pattern instructions only | ~200KB |
| `write_essay_pg` | 51KB — Paul Graham essays as EXAMPLE | Compact style guide | ~50KB |
| `write_micro_essay` | 51KB — same Paul Graham essays | Compact style guide, 250w limit | ~50KB |
| `write_nuclei_template_rule` | 68KB — 1750-line Nuclei docs as CONTEXT | 12-point KEY RULES | ~65KB |

`sanitize_broken_html_to_markdown` intentionally skipped — its 85KB contains load-bearing custom Vitepress styling rules.

### Shell completions

- Added `--show-metadata` to bash, fish, and zsh completions
- Added `--list-transcription-models` to fish completion (function existed but flag was unregistered)

### Maintenance tooling

New script `scripts/audit-patterns.sh` reports:
1. Thin patterns (system.md < 15 lines)
2. Bloated patterns (system.md > 50KB)
3. Stale model/vendor references (GPT-4, ChatGPT, etc.)
4. Patterns missing `# INPUT` section
5. i18n locale key gaps
6. Shell completion gaps vs `flags.go`

```bash
./scripts/audit-patterns.sh          # report only
./scripts/audit-patterns.sh --strict # exit 1 if issues found (CI use)
```

## Test plan

- [ ] `fabric -p improve_prompt` on a prompt returns an improved version
- [ ] `fabric -p explain_math` on "what is the chain rule" returns structured output
- [ ] `fabric -p write_essay_pg` on a topic returns a PG-style essay
- [ ] `fabric -p write_nuclei_template_rule` on a CVE description returns valid YAML
[...]

PR #2122, opened 2026-05-16
feat(patterns): add generate_frontmatter for PKM/Obsidian users

OdinKral:
## Summary

Adds `generate_frontmatter` — a pattern that produces clean, paste-ready YAML frontmatter from any text input.

Designed for PKM workflows (Obsidian, Logseq, or any markdown notes vault), but useful for anyone who wants structured metadata quickly.

## Output fields

| Field | Description |
|---|---|
| `title` | Exact or inferred title |
| `aliases` | 1–3 alternative names to search by |
| `tags` | 3–8 specific lowercase tags, supporting hierarchy (`philosophy/stoicism`) |
| `type` | Document type: article, chapter, transcript, meeting-notes, essay, reference, idea, book |
| `author` | Extracted or omitted |
| `source` | URL or citation if available |
| `date` | Document date or omitted |
| `summary` | One sentence (15–25 words) capturing the core content |
| `status` | Starts as `unprocessed` |

## Usage

```shell
cat article.md     | fabric -p generate_frontmatter
pbpaste            | fabric -p generate_frontmatter
yt URL --transcript | fabric -p generate_frontmatter
```

Output is only the YAML block — paste directly at the top of a note.

## Design notes

- Tags use lowercase + hyphens; hierarchical tags use slash notation
- Fields that can't be inferred are omitted rather than guessed
- Summary is always a complete sentence, not a fragment
- Status defaults to `unprocessed` as the entry point for a processing workflow

Closes #1235.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2170, opened 2026-07-29
fix(patterns): normalize INPUT section markers across 92 patterns (normalization only)

OdinKral:
Resubmit of #2125, untangled from #2126 per @ksylvan's review. This is item 1 of the requested decomposition: **pattern normalization only — no content changes.**

### What this does

Patterns used three different markers for the input section, and most had none at all. This normalizes every one of them to the `# INPUT` heading the majority of the library already uses:

| Before | Count | After |
|---|---|---|
| *(no INPUT marker)* | 79 | appended `# INPUT` + `INPUT:` |
| `## INPUT` | 11 | `# INPUT` (heading level) |
| `**INPUT**` | 1 | `# INPUT` (bold → heading) |
| `**# INPUT:**` | 1 | `# INPUT` (bold → heading) |

92 files, +326 / −14.

### Verifying it is genuinely content-neutral

Rather than ask you to take that on trust, I checked it mechanically: with INPUT markers and trailing whitespace ignored, **every touched file is byte-identical to its previous version.** No prompt text was added, removed, or reworded.

Two fixes versus what #2125 contained:

- **`convert_to_markdown` is now excluded.** It already delimits its input with a bare `INPUT` token inside `<content>` tags. The marker #2125 added landed *inside* those tags, which would have put a stray heading into the content the model is asked to convert. It keeps its existing convention.
- **`create_formal_email` is normalized rather than doubled.** #2125 appended a second marker to a file that already had `**# INPUT:**`. This converts the existing one instead.

Also fixed a CRLF→LF conversion in `create_user_story` that made #2125 show a 45-line rewrite of an unchanged file; it is now a 5-line diff.

### Not in this PR

- Content/editorial changes — held for separate per-pattern PRs (item 2)
- The audit script — #2171 (item 3)

Two patterns in this set (`create_flash_cards`, `generate_code_rules`) are also touched by editorial PRs I'm holding until this one lands, so there are no competing PRs over the same files this time.


OdinKral:
Sorry for the notification burst — posting one map here rather than repeating it on each PR.

Acting on your review of #2125/#2126, plus a second problem I found while doing it: all of my open PRs were on an org-owned fork you can't push to (the #2123 failure). Everything is now on a personal fork and all nine are mergeable.

**Your requested decomposition:**

| PR | | |
|---|---|---|
| #2170 | 1 — normalization only | 92 files, structural only, verified content-neutral |
| #2172 | de-bloat, split out | 216 KB embedded podcast transcript, +2/−6 |
| #2171 | 3 — tooling standalone | the audit script from #2124 |

**Moved to the personal fork, otherwise unchanged** — #2173 (was #2117, also rebased; it had gone `CONFLICTING` after #2167), #2174 (#2118), #2175 (#2119), #2176 (#2120), #2177 (#2121), #2178 (#2122).

**Closed and not resubmitted:** #2124. Two parts of it I'm withdrawing rather than re-filing — the `write_nuclei_template_rule` strip was wrong (those 1,773 lines are the Helper Functions/Extractors reference the prompt tells the model to read), and the `write_essay_pg`/`write_micro_essay` change is a real prompt-engineering tradeoff rather than cleanup, so it shouldn't ride along with mechanical work.

**Deliberately not filed yet:** four per-pattern content PRs (item 2), held until #2170 lands. Two of them touch files #2170 touches, and I'd rather not put competing PRs in your queue again. Happy to open them whenever you want them.

No two open PRs touch the same file.

PR #2171, opened 2026-07-29
feat(scripts): add pattern/maintenance audit script

OdinKral:
Item 3 of @ksylvan's requested decomposition: the maintenance tooling from #2124, on its own, with nothing else attached.

Adds `scripts/audit-patterns.sh` — a dependency-free report over the patterns library and a few adjacent consistency checks:

1. **Thin patterns** — `system.md` under 15 lines (likely stubs)
2. **Bloated patterns** — `system.md` over 50 KB (likely embedded examples)
3. **Stale model refs** — hardcoded GPT-4 / ChatGPT / vendor model names
4. **Missing INPUT** — pattern has no `# INPUT` section
5. **i18n key gaps** — keys in `en.json` missing from other locales
6. **Completion gaps** — long flags in `flags.go` absent from the completions files

It reports only; it never edits patterns. Exits 0 by default so it can be run casually, or `--strict` to exit 1 on findings if you ever want it in CI.

### Sample

```
$ ./scripts/audit-patterns.sh
━━━ Bloated patterns (> 50 KB — likely embedded examples) ━━━
  ⚠  extract_insights_dm  (225 KB)  →  data/patterns/extract_insights_dm/system.md
  ⚠  sanitize_broken_html_to_markdown  (85 KB)  →  data/patterns/sanitize_broken_html_to_markdown/system.md

━━━ Stale model/vendor references ━━━
  ⚠  improve_prompt  →  found: GPT-3 GPT-4 OpenAI's gpt-4
```

Against the current tree it reports 109 issues, which is what motivated the separate normalization and de-bloating PRs.

### Fixed since #2124

`--strict` was parsed as the positional `patterns_dir` argument, so `./audit-patterns.sh --strict` searched a directory named `--strict`, found nothing, and reported **"None found" for every check with exit 0** — an audit tool falsely reporting clean. Now parsed properly, with a fail-loud check if the patterns directory doesn't exist.

Also: colors are emitted only when stdout is a terminal (redirected output was full of raw escape codes), paths print repo-relative instead of `scripts/../data/patterns/...`, and `--help` works.

Tested: default run, `--strict` (exit 1 with findings), `--help`, nonexistent directory (exit 2), unknown flag (exit 2), and redirected output.


ksylvan (approved):
This required GNU grep to be installed (for the scanning of the shell completions gap:

```
━━━ Shell completion gaps (long flags in flags.go missing from completions) ━━━
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
  ✓  All completions in sync
```

And when it was installed, it appeared inaccurate:

```
━━━ Shell completion gaps (long flags in flags.go missing from completions) ━━━
  ⚠  fabric.bash  →  missing --show-metadata
  ⚠  _fabric  →  missing --show-metadata
  ⚠  fabric.fish  →  missing -l list-transcription-models
  ⚠  fabric.fish  →  missing -l show-metadata
```

(accurate for show-metadata, but not for `list-transcription-models`. Nontheless, this is very useful, thank you!


```
rg -l list-transcription-models
docs/Using-Speech-To-Text.md
completions/fabric.bash
completions/fabric.fish
completions/_fabric
internal/cli/help.go
internal/cli/flags.go
```

PR #2177, opened 2026-07-29
fix(completions): add --list-transcription-models and --show-metadata to bash and fish

OdinKral:
Replaces #2121 — same commits, resubmitted from a **personal** fork.

On #2123 you were unable to push maintainer fixes to my branch:

```
ERROR: Permission to AnsibleBBridge/Fabric.git denied to ksylvan.
```

That fork is owned by an organization, and GitHub's "allow edits by maintainers" grant only works for personally-owned forks — the API still reports `maintainerCanModify: true`, so there was no signal from my side that anything was wrong. All of my open PRs were on that fork and had the same problem, so I've moved them to `OdinKral/fabric-contrib`, which is a personal fork. You should be able to push to these.

---

## Summary

Follow-up to #2118 (zsh). The same two flags were also missing from `fabric.bash` and `fabric.fish`:

- `--list-transcription-models` — absent from bash opts string and fish
- `--show-metadata` — absent from bash opts string and fish

Keeps all three completion scripts in sync with the current flag set.

## Changes

- `completions/fabric.bash` — added both flags to the opts string
- `completions/fabric.fish` — added two `complete` lines next to related flags


ksylvan:
Superceded by #2179 which also fixes other completion bugs

PR #2182, opened 2026-07-30
fix(ollama): close stream channel on errors

### LINKED ISSUE #2181: [Bug]: Ollama SendStream leaves goroutine deadlocked when o.client.Chat errors
### What happened?

When using `--stream` with the Ollama vendor, if `o.client.Chat(...)` inside `SendStream` (internal/plugins/ai/ollama/ollama.go) returns an error, the function returns without closing `channel`:

```go
if err = o.client.Chat(ctx, &req, respFunc); err != nil {
    return  // channel never closed here
}
close(channel)
```

`Chatter.Send` in `internal/core/chatter.go` reads that channel with `for update := range responseChan`, which only exits when the channel is closed. Since it never gets closed on this error path, the goroutine blocks forever and the process dies with `fatal error: all goroutines are asleep - deadlock!` instead of surfacing the actual Ollama error.

This is a separate bug from the one fixed in #2061 (which addressed a race on `errChan` writes) — this one is specifically about `responseChan` never being closed on `SendStream`'s early-return paths.

### Expected behavior

The underlying error from `o.client.Chat` (timeout, model unload, connection drop, etc.) should be returned to the caller and printed, not silently swallowed behind a permanent hang.

### Actual behavior

Process panics with a Go runtime deadlock instead of surfacing the real error.

### How did you install Fabric?

go install

### Version

1.4.439 (confirmed still reproducible on latest as of this report)

### Relevant log output

```shell
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive, 5 minutes]:
github.com/danielmiessler/fabric/internal/core.(*Chatter).Send(...)
	internal/core/chatter.go:115 +0x4dc
github.com/danielmiessler/fabric/internal/cli.handleChatProcessing(...)
	internal/cli/chat.go:91 +0x494
```

### Suggested fix

Close `channel` on every return path in `SendStream`, e.g. `defer close(channel)` right after the channel is created, rather than a single `close(channel)` only on the success path. Happy to submit a PR for this.

drawliin:
## Related issues

Closes #2181.

Related context: discussion #2067.

Fixes an Ollama streaming deadlock by making `SendStream` close its response channel on every return path.

Previously, `internal/plugins/ai/ollama.SendStream` only closed the stream channel after `o.client.Chat(...)` completed successfully. If request creation or the Ollama chat call returned an error, `internal/core.Chatter.Send` could block forever while ranging over the unclosed response channel, preventing the original error from being surfaced.

This PR:

- changes Ollama streaming to `defer close(channel)`
- removes the success-only manual close
- adds a regression test that forces an Ollama chat error and verifies the stream channel is closed

## Testing

- `go test ./internal/plugins/ai/ollama`
- `go test ./internal/core -run "TestChatter_SendWithStream|TestChatter_Send_WithStream|TestChatter_Stream|TestRecordFirstStreamError"`


ksylvan (approved):
The fix closes the response channel on each return path. The regression test reproduces the Ollama chat error. I ran the focused tests, the full Go test suite, gofmt, and go vet. All checks passed.

PR #2192, opened 2026-08-12
fix(patterns): patterns loader temp dir leak

### LINKED ISSUE #2190: [Bug]: PatternsLoader leaks one empty temp directory per fabric invocation
`configure()` on `PatternsLoader` calls `os.MkdirTemp` on every plugin registry
construction, so every fabric invocation creates an empty `/tmp/fabric-patterns-*`
directory that is never removed.

Only the pattern download path uses that folder, and it already cleans up after
itself (`movePatterns`, `patterns_loader.go:210`). Every other invocation creates the
directory and never touches it again.

For scale: a daemon here shells out to `fabric` per item and produced 190 empty
directories in 50 minutes on a `tmpfs` `/tmp`. `systemd-tmpfiles` ages `/tmp` at
10 days, so that is roughly 55k directories before the first reap.

## Repro

Requires a configured fabric, meaning an existing `~/.config/fabric/.env`. Contents
are irrelevant, an empty file is enough. It does **not** reproduce on a pristine
HOME, because fabric exits before the plugin registry is built.

```bash
export TMPDIR=$(mktemp -d)
for i in $(seq 20); do fabric -l >/dev/null 2>&1; done
ls -d "$TMPDIR"/fabric-patterns-* | wc -l   # 20, all empty
```

## Scope

- One directory per invocation, on success and on failure, with valid or invalid input.
- `--version` and `--help` are the only exempt commands. They exit before `NewPluginRegistry`.
- Verified on `main` at `338b89c` (= v1.4.470, current release) with a source build,
  and separately with the v1.4.470 release binary.
- Linux amd64 only. There is no OS-specific branch in this path so I expect the same
  on macOS and Windows, but I have not tested those.

## Cause

```
internal/cli/initialization.go:28     core.NewPluginRegistry(fabricDb)
internal/core/plugin_registry.go:107  _ = ret.Configure()
internal/core/plugin_registry.go:454  _ = o.PatternsLoader.Configure()
internal/plugins/plugin.go:107        err = o.ConfigureCustom()
internal/tools/patterns_loader.go:63  os.MkdirTemp("", "fabric-patterns-")
```

`tempPatternsFolder` is read only inside the `PopulateDB()` subtree, and `PopulateDB`
is reached only from `Setup()` and `internal/cli/configuration.go:11` (`-U` / `--setup`).

A plain `defer os.RemoveAll(tempDir)` inside `configure()` would be wrong: it deletes
the folder before `gitCloneAndCopy` populates it. Creating it lazily in `PopulateDB`
is the fix. Happy to send a PR.

## Environment

- Operating System: Linux - amd64
- OS Version: `Linux 7.0.0-28-generic x86_64`, Ubuntu 26.04 LTS
- How installed: From Source (also reproduced with the Linux amd64 release binary)
- Version: `v1.4.470`

scottidler:

Stops `PatternsLoader` from leaking one empty temp directory per fabric invocation.

`configure()` called `os.MkdirTemp` directly. It is wired as the plugin's
`ConfigureCustom`, so it runs on every `NewPluginRegistry` construction, which means
every fabric invocation created an empty `fabric-patterns-*` directory that nothing
removed. Details and measurements are in #2190.

Only `PopulateDB()` and the helpers it calls read `tempPatternsFolder`, and the
success path already removes the folder in `movePatterns`. So this moves creation into
`ensureTempPatternsFolder()`, called from `PopulateDB()`, plus a deferred cleanup that
also covers the error paths and leaves the loader reusable.

Worth noting for review: a plain `defer os.RemoveAll(tempDir)` inside `configure()`
looks like the obvious one-line fix but is wrong. It deletes the folder before
`gitCloneAndCopy` populates it. Lazy creation is the correct seam.

### Verification

- 20 invocations of `fabric -l` against a configured `~/.config/fabric/.env`:
  20 leaked directories before, 0 after.
- `fabric -U` behaves identically before and after: 255 patterns downloaded,
  257 entries installed, `loaded` marker written, 0 leftover temp directories.
- `go test ./internal/tools/... ./internal/core/... ./internal/plugins/...` passes,
  25 packages, exit 0. `gofmt` and `go vet` clean.
- The new tests fail against unpatched code, confirming they cover the regression
  rather than just passing alongside it.
- Verified on Linux amd64, source build of `main` at `338b89c`.

## Related issues

closes #2190

## Screenshots

N/A, no user-visible output changes.

### DIFF
--- a/internal/tools/patterns_loader.go
+++ b/internal/tools/patterns_loader.go
@@ -59,9 +59,21 @@ type PatternsLoader struct {
 
 func (o *PatternsLoader) configure() (err error) {
 	o.pathPatternsPrefix = fmt.Sprintf("%v/", o.DefaultFolder.Value)
+	return
+}
+
+// ensureTempPatternsFolder lazily creates the scratch folder used by the pattern
+// download path. Only PopulateDB and the helpers it calls need it, so it must not
+// be created in configure(): configure() runs on every plugin registry
+// construction, which leaked one empty temp directory per fabric invocation.
+func (o *PatternsLoader) ensureTempPatternsFolder() (err error) {
+	if o.tempPatternsFolder != "" {
+		return
+	}
+
 	// Use a consistent temp folder name regardless of the source path structure
-	tempDir, err := os.MkdirTemp("", "fabric-patterns-")
-	if err != nil {
+	var tempDir string
+	if tempDir, err = os.MkdirTemp("", "fabric-patterns-"); err != nil {
 		return fmt.Errorf(i18n.T("patterns_failed_create_temp_folder"), err)
 	}
 	o.tempPatternsFolder = tempDir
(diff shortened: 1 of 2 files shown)


ksylvan:
Closing this - superceded by #2211 - Thank you for your contribution, @scottidler

PR #2198, opened 2026-08-26
feat(providers): add Synthorai as an OpenAI-compatible provider

cuihuan:
**Disclosure: I'm affiliated with Synthorai** — flagging that this adds our own product.

Synthorai (https://synthorai.io) is an OpenAI-compatible gateway: one key and base URL reach Anthropic, OpenAI, Google, Z.ai, Moonshot, DeepSeek and Qwen models, billed at each upstream provider's own list price. Docs: https://synthorai.io/docs

Same three-line shape as the existing Novita AI / OpenRouter / SiliconCloud entries in `ProviderMap`, plus the provider list in both READMEs.

## Changes
- `internal/plugins/ai/openai_compatible/providers_config.go` — one `ProviderMap` entry
- `README.md` / `README.zh.md` — added to the provider list, keeping both in sync

## Notes

`ImplementsResponses: false`, matching every other gateway in that map. Synthorai does serve an OpenAI Responses endpoint, but I haven't exercised fabric's Responses code path against it, so I left the conservative value rather than assert a capability I hadn't tested here.

Verified before opening: `GET https://synthorai.io/v1/models` and `POST /v1/chat/completions` both return 200 against production, so `--listmodels` and normal chat should work out of the box. `go build ./...`, package tests and `gofmt` all clean.

There's a free tier (10 calls, up to $1, no card) if you'd like to try it.


cuihuan:
Small ask: **could a maintainer approve the workflow run on this PR?**

It currently shows zero checks. `ci.yml` has `paths-ignore` for `**/*.md` and `data/patterns/**`, but this PR also touches `internal/plugins/ai/openai_compatible/providers_config.go`, so `Run tests` should fire — it looks like the first-time-contributor approval gate rather than a path skip. (A comparison PR opened after this one shows `Run tests :: success`.)

While you're here, the diff is 5 lines in `ProviderMap`, placed alphabetically between `SiliconCloud` and `Together`, in the same shape as the neighbouring entries:

```go
"Synthorai": {
    Name:                "Synthorai",
    BaseURL:             "https://synthorai.io/v1",
    ImplementsResponses: false,
},
```

**On `ImplementsResponses: false`** — worth stating since it's the one judgement call here. We do publish a `/v1/responses` surface, but I've only actually verified `POST /v1/chat/completions`, `POST /v1/messages` and `GET /v1/models` returning 200 against a real key. I haven't exercised `/v1/responses` end to end, and a 401 from it proves nothing because our auth sits on the `/v1` prefix — unrouted paths 401 too. So `false` is the conservative value rather than the accurate-sounding one. If you'd rather it were `true`, tell me and I'll verify it properly first and come back with the result.

**Disclosure:** I'm affiliated with Synthorai — vendor adding its own entry, not a user request, and I'm not claiming demand.

PR #2203, opened 2026-08-29
fix(web): preserve multi-byte UTF-8 split across SSE chunks in chat stream

pacocartones:
## What & why

`ChatService.createMessageStream()` decoded each network chunk with a fresh `new TextDecoder().decode(value)` and **no `{ stream: true }`**. Network chunk boundaries don't respect UTF-8 character boundaries, so a multi-byte rune (CJK, emoji) split across two reads is decoded as two incomplete halves and corrupted into `U+FFFD` — permanently baked into `buffer` before `JSON.parse`. The `\n\n` frame buffering is correct; the corruption is one layer below, at byte decode.

The server ships raw multi-byte UTF-8 (`internal/server/chat.go` writes `data: %s\n\n` from `json.Marshal`, which keeps CJK/emoji as raw bytes), so real LLM tokens routinely trigger this.

## Fix

Hoist a single persistent `TextDecoder` out of the read loop and decode with `{ stream: true }`, so pending bytes carry across chunks.

## Verification

Replaying the exact transform with a stream split inside a 4-byte emoji (`F0 9F | 8C 8D`):

```
server sent:          "你好世界 🌍"
before (HEAD):        "你好世界 ���"   (contains U+FFFD)
after  (this change): "你好世界 🌍"    (intact)
```


ksylvan (approved):
@pacocartones Thanks for this. I found the same fix needed to be made in a couple of other places.

PR #2214, opened 2026-09-04
fix: always send user message with patterns for OpenAI-compatible providers

### LINKED ISSUE #2208: [Bug]: Pattern system prompt merges user input into single system message, causing empty responses on OpenAI-compatible providers that require a user turn
### What happened?

When using a pattern with an OpenAI-compatible provider (e.g. Ollama Cloud via OPENAI_API_BASE_URL), Fabric sends only a single message with role=system containing both the pattern content and the user input appended at the # INPUT: section. No role=user message is sent. Several OpenAI-compatible providers require at least one user turn to generate a response and return an empty response when only a system message is present.

To reproduce

Configure Fabric with an OpenAI-compatible provider via OPENAI_API_BASE_URL (e.g. Ollama Cloud at https://ollama.com/v1)
Run any pattern: echo "test input" | fabric -p summarize
Observe empty response

Enabling --debug=4 confirms only one message is sent:

DEBUG: FABRIC->LLM request messages (1)
DEBUG: FABRIC->LLM [0] role=system content="# IDENTITY and PURPOSE\n...\n# INPUT:\n\nINPUT:\ntest input"
DEBUG: LLM->FABRIC response content=""

The same provider works correctly when called directly with separate system and user messages.

Root cause

In internal/core/chatter.go, BuildSession sets inputUsed = true whenever a pattern is loaded, then uses this flag to suppress appending the user message:

if len(request.Message.MultiContent) > 0 || (request.Message != nil && !inputUsed) {
    session.Append(request.Message)
}

When inputUsed is true the user message is never appended, resulting in a single system message.

What I did to fix it for me but that might not be done because other providers

Remove the !inputUsed guard so the user message is always appended as a separate turn:

go
if len(request.Message.MultiContent) > 0 || request.Message != nil {
    session.Append(request.Message)
}

And remove the now-unused inputUsed variable and assignment.

### Operating System

Linux - amd64

### OS Version

```shell
Debian 13
```

### How did you install Fabric?

Release Binary - Linux (amd64)

### Version

```text
Fabric version: v1.4.473
```

### Relevant log output

```shell

```

### Relevant screenshots (optional)

_No response_

putramkti:

When a pattern is active, `BuildSession` used to suppress the user message entirely because the user's input was already embedded in the system message through the `{{input}}` placeholder. That left the session with a single `role=system` message.

OpenAI-compatible providers like Ollama Cloud (`OPENAI_API_BASE_URL`) require at least one `role=user` turn to produce a response. Without it, they return empty responses.

The fix removes the `inputUsed` guard so the user message is always appended as a separate turn. The input still appears in the system message via the pattern (nothing changes there), and now the user message is also present so providers that need it get one.

## Related issues

Closes #2208

## How to test

1. Configure Fabric with an OpenAI-compatible provider:
   ```bash
   export OPENAI_API_BASE_URL=https://ollama.com/v1
   ```
2. Run any pattern:
   ```bash
   echo "The quick brown fox" | fabric -p summarize
   ```
3. Before this fix: empty response. After: normal summary output.

For unit tests:
```bash
go test ./internal/core/ -run TestBuildSession_PatternAlwaysSendsUserMessage -v
```

## Notes

The input appears twice when a pattern is loaded (once in the system message, once as the user message). This is intentional. Most models handle it fine, and it matches what most OpenAI wrappers do when combining a system prompt with user input. If a provider ever complains about duplicated content, that's a separate issue to address with per-provider message shaping.

The existing test `TestChatter_BuildSession_SeparatesSystemSections` was updated to expect 2 messages (system + user) instead of 1, matching the new behavior.

### DIFF
--- a/internal/core/chatter.go
+++ b/internal/core/chatter.go
@@ -261,7 +261,7 @@ func (o *Chatter) BuildSession(request *domain.ChatRequest, raw bool) (session *
 	}
 
 	var patternContent string
-	inputUsed := false
+
 	if request.PatternName != "" {
 		var pattern *fsdb.Pattern
 		if request.NoVariableReplacement {
@@ -274,7 +274,7 @@ func (o *Chatter) BuildSession(request *domain.ChatRequest, raw bool) (session *
 			return nil, fmt.Errorf(i18n.T("chatter_error_get_pattern"), request.PatternName, err)
 		}
 		patternContent = pattern.Pattern
-		inputUsed = true
+
 	}
 
 	systemMessage := joinPromptSections(contextContent, patternContent)
@@ -338,9 +338,9 @@ func (o *Chatter) BuildSession(request *domain.ChatRequest, raw bool) (session *
 		if systemMessage != "" {
 			session.Append(&chat.ChatCompletionMessage{Role: chat.ChatMessageRoleSystem, Content: systemMessage})
 		}
-		// If multi-part content, it is in the user message, and should be added.
-		// Otherwise, we should only add it if we have not already used it in the systemMessage.
-		if len(request.Message.MultiContent) > 0 || (request.Message != nil && !inputUsed) {
+		// Always append the user message so that OpenAI-compatible providers
+		// that require at least one user turn receive one.
+		if request.Message != nil {
 			session.Append(request.Message)
 		}
 	}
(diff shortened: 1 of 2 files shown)


putramkti:
Friendly ping, this is ready for review. Happy to adjust anything you want changed.

PR #2216, opened 2026-09-06
fix: enable raw mode for GPT-6 models

### LINKED ISSUE #2215: Codex gpt-6 models fail with unsupported temperature parameter
## Description

Codex models in the `gpt-6` family are not automatically put into raw mode, so Fabric sends sampling parameters that the Codex backend rejects.

## Reproduction

Using Fabric v1.4.455 (also confirmed that v1.4.477 still has the same `NeedsRawMode` prefix list):

```sh
printf 'Reply only OK.\n' | fabric -V Codex -m gpt-6-astra
```

Fabric reports:

```text
codex request failed with status 400
```

The backend response body is:

```json
{"detail":"Unsupported parameter: temperature"}
```

Running the same request with `--raw` succeeds:

```sh
printf 'Reply only OK.\n' | fabric -V Codex -m gpt-6-astra --raw
# OK
```

## Cause

`internal/plugins/ai/openai/openai.go` enables raw mode for the `gpt-5` prefix but not `gpt-6`. Consequently the request contains `temperature: 0.7` and `top_p: 0.9`.

## Suggested fix

Add `gpt-6` to the prefixes in `NeedsRawMode`, with a regression test.

ctbaum:
## Summary

- enable automatic raw mode for `gpt-6` models
- add a regression test for `gpt-6-astra`

Without raw mode, Fabric sends `temperature` and `top_p`; the Codex backend rejects the request with HTTP 400 (`Unsupported parameter: temperature`).

Fixes #2215.

## Testing

```sh
go test ./...
```

### DIFF
--- a/cmd/generate_changelog/incoming/2216.txt
+++ b/cmd/generate_changelog/incoming/2216.txt
@@ -1,0 +1,3 @@
+### PR [#2216](https://github.com/danielmiessler/Fabric/pull/2216) by [ctbaum](https://github.com/ctbaum): fix: enable raw mode for GPT-6 models
+
+- Fix: enable raw mode for GPT-6 models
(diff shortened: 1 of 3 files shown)


ksylvan (approved):
Thanks @ctbaum - Approved

PR #2220, opened 2026-09-17
feat(providers): add Y-API as an OpenAI-compatible provider

jiweiyeah:

Adds **Y-API** as an OpenAI-compatible provider.

Y-API is a single OpenAI-compatible endpoint (`https://api.y-api.bestvirtualgoods.com/v1`) that fronts 15 models from DeepSeek, Z.ai, Moonshot, Tencent, Xiaomi, Qwen and OpenAI behind one key. It works in Fabric today as a custom OpenAI base URL; this entry is so it shows up in `fabric --listvendors` and in `--setup` instead of having to be typed in by hand.

Three files, same shape as the recent provider additions (#2198 Synthorai, #2210 Pzero):

| File | Change |
| --- | --- |
| `internal/plugins/ai/openai_compatible/providers_config.go` | one `"Y-API"` entry in `ProviderMap` |
| `README.md` | bullet in the OpenAI-Compatible Providers list |
| `README.zh.md` | same, in the Chinese list |

Two notes on the entry:

- **The key env var comes out as `Y_API_API_KEY`.** That is derived, not chosen — `BuildEnvVariablePrefix("Y-API")` maps `-` to `_`, then `AddSetting("API Key")` appends `API_KEY`. If you would rather it read `YAPI_API_KEY`, renaming the vendor to `"YAPI"` is the only way, and I am happy to do that instead.
- **`ImplementsResponses: false`.** The endpoint does also serve `POST /v1/responses`, but I set `false` to match the other 20 entries rather than assert a capability I could not re-test this session. Easy to flip if you want it on.

## Related issues

No tracking issue.

## Screenshots

`fabric --listvendors`, built from this branch:

```
Venice AI
Y-API
Z AI
```

### Testing

```
$ go test ./...
# 34 packages ok, no failures

$ gofmt -l internal/plugins/ai/openai_compatible/
# empty

$ go vet ./internal/plugins/ai/openai_compatible/
# clean
```

`internal/plugins/ai/openai_compatible` and `internal/cli` both pass. The one thing I could not run is `nix flake check` — no Nix on this machine — so the formatting gate is covered by `gofmt` only.

I will add `cmd/generate_changelog/incoming/<PR number>.txt` in a follow-up commit on this branch once the PR number exists.


ksylvan (approved):

ksylvan (approved):
Thank you for your contribution @jiweiyeah

